PR #60, opened 2026-01-21
fix(circuit-breaker): Don't count 4xx client errors as failures

Marviel:
## Summary

Client errors (400, 401, 403, 404, etc.) are not indicative of worker health issues - they indicate bad client requests. The circuit breaker should only trip on server errors (5xx) which indicate actual problems with the worker.

## Problem

The regular router in `router.rs` was counting 4xx responses as circuit breaker failures:

```rust
worker.record_outcome(response.status().is_success());  // 4xx counted as failure!
```

This causes healthy workers to be marked unavailable when clients send malformed requests (e.g., `max_tokens` exceeding model context length → 400 Bad Request).

## Solution

Align with the PD router (`pd_router.rs` line 888) which already handles this correctly:

```rust
let status = response.status();
worker.record_outcome(status.is_success() || status.is_client_error());
```

## Changes

1. **`src/routers/http/router.rs`**: Fixed circuit breaker outcome recording
2. **`py_test/integration/test_circuit_breaker.py`**: Added `test_circuit_breaker_ignores_4xx_client_errors` to verify the fix

## Testing

All 6 circuit breaker tests pass:
- `test_circuit_breaker_opens_and_recovers` ✅
- `test_circuit_breaker_half_open_failure_reopens` ✅  
- `test_circuit_breaker_disable_flag` ✅
- `test_circuit_breaker_per_worker_isolation` ✅
- `test_circuit_breaker_with_retries` ✅
- `test_circuit_breaker_ignores_4xx_client_errors` ✅ (new)

## Real-World Impact

This bug caused production outages where clients sending oversized requests (400 errors) would trip the circuit breaker and mark healthy vLLM workers as unavailable.

### DIFF
--- a/py_test/integration/test_circuit_breaker.py
+++ b/py_test/integration/test_circuit_breaker.py
@@ -189,3 +189,49 @@ def test_circuit_breaker_with_retries(router_manager, mock_workers):
         timeout=5,
     )
     assert r.status_code == 200
+
+
+@pytest.mark.integration
+def test_circuit_breaker_ignores_4xx_client_errors(router_manager, mock_workers):
+    """
+    4xx client errors (e.g., 400 Bad Request) should NOT count as circuit breaker
+    failures. Only 5xx server errors indicate worker health issues.
+
+    This test ensures that many 400 responses do not trip the circuit breaker,
+    while 500 responses still do.
+    """
+    # Worker that returns 400 Bad Request for all requests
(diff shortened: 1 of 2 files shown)


Marviel:
@Prowindy hi! Hit this case in our inference setup and I believe it's a bug.

Feel free to correct my understanding if you believe I'm "holding it wrong".

Otherwise, care to review and approve?

PR #70, opened 2026-01-26
Fix cache aware policy for direct vllm-pd-router instance, add OpenTelemetry context propagation

MichaelWang237:
## Purpose

Currently, when using vllm-pd-disaggregation mode, the cache aware policy does not work correctly. This is due to two main reasons:
* No prefix tree is created when the vllm-pd-router is created.
* The load of each worker is not tracked -- so any load-balancing thresholds will fail to apply.

This PR addresses these issues for the direct vLLM router instantiation as follows (i.e. service discovery mode not yet supported):
* A prefix tree is created for both prefill and decode policies using the init_workers() method. In case future stateful routing policies may require initialization, I've refactored init_workers to be part of the LoadBalancingPolicy trait, instead of a method exclusive to the cache aware policy.
* `increment_load()` and `decrement_load()` methods are called inside the `process_vllm_two_stage` function for the prefill/decode worker when they start/finish running their phase of the request.

Other changes:
* I've added some header propagation so OpenTelemetry trace context can be propagated to worker nodes. The current implementation is somewhat hardcoded, and may need to be reworked, but I thought it might be helpful.
* Changed the request payload printouts to `debug` level (previously at `info`) to avoid cluttering the logs.

## Test Plan
Observe routing of LLM calls when:
* Running an open-source coding agent (mini-SWE-Agent) to test the routing. For each request/task it performs, this agent exhibits significant prefix cache sharing in its LLM calls. LLM calls made for handling different tasks do not share prefix with each other. For the same task, we should expect to see all of its LLM calls being handled by the same workers.
* Using a script to spam LLM calls of a specific prefix to the router. We should expect to see the load balancing thresholds kick in.

## Test Result
Arguments:
```
cargo run -- \
    --policy cache_aware \
    --vllm-pd-disaggregation \
    --prefill http://sprc00... \
    --decode http://sprc02...\
    --decode http://sprc03...\
    --host 0.0.0.0 \
    --port 8100 \
    --balance-abs-threshold 5 \
    --balance-rel-threshold 1.5
```
Setting: 3 nodes (named sprc00, sprc02, sprc03), each node runs one instance of vLLM. sprc00 is designated as prefill, and the others are designated as decode nodes (1P2D). Cache aware policy is used for both prefill and decode.

**Before the changes**, we get the following warning for prefill/decode policies:
`vllm_router_rs::policies::cache_aware: src/policies/cache_aware.rs:306: Warning: No tree found for model 'default', using random worker selection`. Running the router confirms this behavior. For instance, running the coding agent shows that a single task is being handled by multiple decode workers, when the high prefix cache hit rate (>90%) between consecutive LLM calls suggests that only one decode worker should handle it.
image

**After the changes**, the router behaves as expected for both scenarios outlined in the test plan. Specifically, for the coding agent, we see that each task is routed to only one (prefill, decode) pair -- (please ignore the errors, they are exceptions raised in the agent's logic during normal operation):
image

For the load-balance threshold testing, we can observe that the thresholds are being applied -- once the sprc02 decode worker has 6+ more in-flight requests than the sprc03 decode worker, sprc03 is chosen, even though the prefix cache hit rate is higher on the former.

image

---
<details>

### DIFF
--- a/src/routers/header_utils.rs
+++ b/src/routers/header_utils.rs
@@ -85,12 +85,13 @@ pub fn propagate_headers(
 ) -> reqwest::RequestBuilder {
     if let Some(h) = headers {
         for (k, v) in h.iter() {
-            if header_names.iter().any(|&name| k.as_str().eq_ignore_ascii_case(name)) {
+            if header_names
+                .iter()
+                .any(|&name| k.as_str().eq_ignore_ascii_case(name))
+            {
                 request = request.header(k, v);
             }
         }
     }
     request
 }
-
-
--- a/src/routers/http/vllm_pd_router.rs
+++ b/src/routers/http/vllm_pd_router.rs
@@ -298,8 +298,6 @@ impl VllmPDRouter {
         }
     }
 
-
-
     /// Two-stage request processing for vLLM disaggregated mode using discovered endpoints
     async fn process_vllm_two_stage_request_discovered(
         &self,
@@ -384,7 +382,8 @@ impl VllmPDRouter {
             .header("X-Request-Id", &request_id); // P2P coordination metadata in header
 
         // Propagate trace headers and add X-data-parallel-rank header using shared utilities
-        prefill_request_builder = header_utils::propagate_trace_headers(prefill_request_builder, headers);
+        prefill_request_builder =
+            header_utils::propagate_trace_headers(prefill_request_builder, headers);
         prefill_request_builder =
             dp_utils::add_dp_rank_header(prefill_request_builder, prefill_dp_rank);
         if let Some(rank) = prefill_dp_rank {
@@ -475,7 +474,8 @@ impl VllmPDRouter {
             .header("X-Request-Id", &request_id); // Same P2P coordination metadata in header
 
         // Propagate trace headers and add X-data-parallel-rank header using shared utilities
-        decode_request_builder = header_utils::propagate_trace_headers(decode_request_builder, headers);
+        decode_request_builder =
+            header_utils::propagate_trace_headers(decode_request_builder, headers);
         decode_request_builder =
             dp_utils::add_dp_rank_header(decode_request_builder, decode_dp_rank);
         if let Some(rank) = decode_dp_rank {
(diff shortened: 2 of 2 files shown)


MichaelWang237:
Hi @Prowindy, thanks for your time and feedback! I've changed the `info!`s to `debug!`s and have refactored the trace propagation code into a helper function in `header_utils.rs`.

It seems the buildkite failure is due to an issue with the test environment -- there appears to be a missing pip package (see log snippet below) -- it doesn't seem like it's related to the router itself. Could you take a look on your side?

Also, are there any documents/resources for staying up to date with the future plans/directions of the project? I'm interested in getting involved with this project (as time allows) and would like my contributions to be as useful as possible. Thanks!

```
[2026-01-28T00:26:47Z] ✓ Connectivity test passed
--
[2026-01-28T00:26:47Z]
[2026-01-28T00:26:47Z] ℹ Running LM-Eval accuracy test on gsm8k task...
[2026-01-28T00:26:47Z] ℹ This may take several minutes...
[2026-01-28T00:26:52Z] ✗ LM-Eval failed: Fast download using 'hf_transfer' is enabled (HF_HUB_ENABLE_HF_TRANSFER=1) but 'hf_transfer' package is not available in your environment. Try `pip install hf_transfer`.
[2026-01-28T00:26:52Z] ✗ Evaluation failed: Fast download using 'hf_transfer' is enabled (HF_HUB_ENABLE_HF_TRANSFER=1) but 'hf_transfer' package is not available in your environment. Try `pip install hf_transfer`.
[2026-01-28T00:26:52Z] + cleanup_instances
[2026-01-28T00:26:52Z] + echo '=== Cleaning up vLLM instances ==='
[2026-01-28T00:26:52Z] === Cleaning up vLLM instances ===
[2026-01-28T00:26:52Z] + pkill -f 'vllm serve'
[2026-01-28T00:26:52Z] + [[ -n 11978 ]]
[2026-01-28T00:26:52Z] + echo '=== Cleaning up router ==='
[2026-01-28T00:26:52Z] === Cleaning up router ===
[2026-01-28T00:26:52Z] + kill 11978
[2026-01-28T00:26:52Z] + sleep 2
[2026-01-28T00:27:03Z] 🚨 Error: The command exited with status 1
[2026-01-28T00:27:03Z] user command error: The plugin docker command hook exited with status 1
```

PR #81, opened 2026-02-06
feat: Aggregate models from all workers in /v1/models endpoint

surak:
## Summary

The `/v1/models` endpoint now aggregates model lists from all registered workers, providing a complete view of all available models across the router's worker pool.

## Changes

- Add `aggregate_models()` method to `Router` that fetches `/v1/models` from all workers in parallel
- Add `aggregate_models()` method to `RouterManager` for IGW (multi-model) mode
- Combine and deduplicate model lists by ID
- Returns complete OpenAI-style model response with full metadata (id, object, created, owned_by, root, max_model_len, permission)

## Motivation

When running vLLM workers with multiple `--served-model-name` aliases (e.g., serving the same model under different names like `gpt-3.5-turbo`, `text-davinci-003`, etc.), each worker returns its own model list. The router now aggregates these to provide a unified `/v1/models` response.

This is particularly useful when:
- Workers have different model aliases for the same underlying model
- Multiple workers serve overlapping model sets
- Users need to discover all available models via the router

## Testing

Tested with two workers:
- Worker on port 8000: 2 models (alias-apertus, "15 - Apertus-8B-Instruct-2509...")
- Worker on port 8005: 5 models (alias-embeddings, gpt-3.5-turbo, text-davinci-003, text-embedding-ada-002, alias-qwen3-8b-embeddings)

Router `/v1/models` now returns all 7 models with complete metadata.


Prowindy (commented):
the motivation makes sense. can we fix buildkite?

PR #82, opened 2026-02-08
fix(consistent-hash): pass HTTP headers to policy in PD disaggregation mode

RGB-loop:
## Purpose

In PD disaggregation mode (`--pd-disaggregation` and `--vllm-pd-disaggregation`), the `consistent_hash` policy's HTTP header-based session affinity (e.g., `X-Session-ID`) was silently broken. Headers were never passed to the policy because `PDRouter` and `VllmPDRouter` called `select_worker()` instead of `select_worker_with_headers()`.

The `LoadBalancingPolicy` trait already provides full header support via `select_worker_with_headers()`, and `ConsistentHashPolicy` correctly implements header extraction. The regular `Router` (non-PD mode) was already calling `select_worker_with_headers()` correctly — only the PD routers were missing this.

This is the same class of bug as #70 (cache_aware policy not properly wired in PD mode).

## Changes

- `PDRouter.pick_worker_by_policy_arc`: accept `RequestHeaders` and call `select_worker_with_headers` instead of `select_worker`
- `PDRouter.select_pd_pair`: accept `HeaderMap`, convert to `RequestHeaders`, and pass through
- `VllmPDRouter.select_worker_with_policy`: accept `RequestHeaders` and call `select_worker_with_headers`
- `VllmPDRouter` route_chat/route_completion/transparent_proxy: convert headers and pass to policy selection in all code paths
- Add 3 unit tests for PD mode header-based consistent hash routing

Other policies (`random`, `round_robin`, `power_of_two`, `cache_aware`) are unaffected as they ignore the headers parameter (`_headers`).

## Test Plan

- [x] Added unit tests: `test_select_worker_pair_with_headers_pd_mode`, `test_pd_mode_header_takes_priority_over_body`, `test_pd_mode_without_headers_falls_back_to_body`
- [x] Manual test: PD mode + `--policy consistent_hash` + `X-Session-ID` header routes consistently to same worker pair

### DIFF
--- a/src/routers/http/pd_router.rs
+++ b/src/routers/http/pd_router.rs
@@ -8,7 +8,7 @@ use crate::core::{
     WorkerFactory, WorkerLoadGuard, WorkerRegistry, WorkerType,
 };
 use crate::metrics::RouterMetrics;
-use crate::policies::{LoadBalancingPolicy, PolicyRegistry};
+use crate::policies::{LoadBalancingPolicy, PolicyRegistry, RequestHeaders};
 use crate::protocols::spec::{
     ChatCompletionRequest, ChatMessage, CompletionRequest, GenerateRequest, RerankRequest,
     ResponsesRequest, StringOrArray, UserMessageContent,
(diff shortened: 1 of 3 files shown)


Prowindy:
@RGB-loop could you please rerun test on top of https://github.com/vllm-project/router/pull/85, which fixed a partial problem?

PR #93, opened 2026-02-19
fix: make prepare_prefill_request works for generate API

wpc:
The generate API (inference/v1/generate) places max_tokens and min_tokens inside a nested sampling_params object, unlike OpenAI-style endpoints (chat/completions) which use top-level fields. Previously, prepare_prefill_request only patched top-level fields, which meant prefill requests to the generate API were not correctly capped at max_tokens=1.

This change adds a path parameter to prepare_prefill_request and uses it to select the correct patching strategy:
- For inference/v1/generate: patch sampling_params.max_tokens and sampling_params.min_tokens
- For other endpoints: patch top-level max_tokens, max_completion_tokens, and min_tokens (existing behavior)

Also deduplicates inline prefill request preparation in process_vllm_two_stage_request_discovered by reusing the shared prepare_prefill_request method.

PLEASE FILL IN THE PR DESCRIPTION HERE ENSURING ALL CHECKLIST ITEMS (AT THE BOTTOM) HAVE BEEN CONSIDERED.

## Test Result

---
<details>

- [ ] The purpose of the PR, such as "Fix some issue (link existing issues this PR will resolve)".
- [ ] The test plan, such as providing test command.
- [ ] The test results, such as pasting the results comparison before and after, or e2e results

### DIFF
--- a/src/routers/http/vllm_pd_router.rs
+++ b/src/routers/http/vllm_pd_router.rs
@@ -140,6 +140,7 @@ impl VllmPDRouter {
     /// Modify request for prefill stage (set max_tokens=1)
     /// - For inference/v1/generate: patch sampling_params.max_tokens and sampling_params.min_tokens
     /// - For other endpoints (fallback): patch top-level max_tokens, max_completion_tokens, min_tokens
+    ///
     /// stream=false and stream_options removal are always applied at top level.
     fn prepare_prefill_request(mut request: Value, path: &str) -> Value {
         if path.contains("inference/v1/generate") {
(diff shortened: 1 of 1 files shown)


Prowindy:
A better fix may be in vLLM to make max_tokens and min_tokens at top-level be respected in /generate handling, rather than duplicating another max_tokens and min_tokens in sampling_params. cc @ywang96 @simon-mo 

But since that approach involves a non-trivial migration for existing /generate users, we can unblock this on router level for now.

PR #96, opened 2026-02-22
Fix IPv6 worker URL formatting

aarkharov:
## Summary
- Wrap IPv6 addresses in brackets per RFC 2732 when constructing worker URLs in `PodInfo::worker_url()`
- Without this fix, service discovery produces invalid URLs like `http://2803:6086:...:8000` instead of `http://[2803:6086:...]:8000`
- This breaks health checks and request forwarding on IPv6-only clusters

## Changes
- `src/service_discovery.rs`: Detect IPv6 addresses (contain `:`) and wrap in `[brackets]`
- Added unit tests

## Test plan
- cargo test --lib service_discovery::tests
- built a new image and verified in an IPv6-only cluster

### DIFF
--- a/src/service_discovery.rs
+++ b/src/service_discovery.rs
@@ -167,9 +167,14 @@ impl PodInfo {
         self.is_ready && self.status == "Running"
     }
 
-    /// Generates a worker URL for this pod
+    /// Generates a worker URL for this pod.
+    /// IPv6 addresses are wrapped in brackets per RFC 2732 (e.g. `http://[::1]:8000`).
     pub fn worker_url(&self, port: u16) -> String {
-        format!("http://{}:{}", self.ip, port)
+        if self.ip.contains(':') {
+            format!("http://[{}]:{}", self.ip, port)
+        } else {
+            format!("http://{}:{}", self.ip, port)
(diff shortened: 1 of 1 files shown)


src/service_discovery.rs
```diff
@@ -167,9 +167,14 @@ impl PodInfo {
...
-    /// Generates a worker URL for this pod
+    /// Generates a worker URL for this pod.
+    /// IPv6 addresses are wrapped in brackets per RFC 2732 (e.g. `http://[::1]:8000`).
     pub fn worker_url(&self, port: u16) -> String {
-        format!("http://{}:{}", self.ip, port)
+        if self.ip.contains(':') {
+            format!("http://[{}]:{}", self.ip, port)
+        } else {
+            format!("http://{}:{}", self.ip, port)
+        }
```
amrmahdi: Use standard network libraries to deal with IPs 

```suggestion
    let addr: IpAddr = ip.parse().expect("pod IP must be valid");
    format!("http://{}", SocketAddr::new(addr, port))
```

You will need to import `std::net::{IpAddr, SocketAddr}`

PR #105, opened 2026-03-06
fix: use parse_worker_url() to strip @rank before reqwest dispatch in VllmPDRouter

### LINKED ISSUE #100: [Bug]: Regression in #98: hostname:port URLs with --intra-node-data-parallel-size > 1 fail with endpoint_url() fix
## Bug Description

Commit `d13949d` (#98, "fix rustfmt and apply IPv6+DP URL fix to VllmPDRouter") introduces a regression when `@rank`-suffixed URLs are constructed from hostname:port inputs when `--intra-node-data-parallel-size > 1`, causing userinfo parsing. The previous commit `bbb3f96` works correctly.

## Environment

- **Platform**: NVIDIA GB200 cluster (aarch64)
- **Router version**: latest main (`d13949d`)
- **vLLM**: nightly-aarch64
- **Mode**: PD disaggregation with NIXL connector

## Steps to Reproduce

Launch router with hostname-based URLs and `--intra-node-data-parallel-size 4`:

```bash
cargo run --release -- \
    --policy round_robin \
    --vllm-pd-disaggregation \
    --max-concurrent-requests 9216 \
    --prefill http://node1:8087 \
    --prefill http://node2:8087 \
    --prefill http://node3:8087 \
    --prefill http://node4:8087 \
    --prefill http://node5:8087 \
    --prefill http://node6:8087 \
    --decode http://node7:8087 \
    --decode http://node8:8087 \
    --decode http://node9:8087 \
    --decode http://node10:8087 \
    --host 0.0.0.0 \
    --port 8123 \
    --intra-node-data-parallel-size 4
```
## Expected Behavior

Router expands each URL into 4 DP-rank endpoints and correctly routes requests to the original hostname.

## Actual Behavior

The `@rank` suffix causes the URL to be parsed as HTTP userinfo (`user:pass@host`). For example:
`http://node1:8087@1/v1/completions` is parsed as:
- **username**: `node1`
- **password**: `8087`
- **host** becomes `1`, and the client attempts to connect to `0.0.0.1`

Router log:

Image

Image

## Potential Root Cause Analysis

In the old code (`bbb3f96`), `VllmPDRouter` used `dp_utils::extract_dp_rank()` to manually parse and strip the `@rank` suffix from URLs before making HTTP requests:

```rust
// Old code
let (prefill_base_url, prefill_dp_rank) = if self.intra_node_data_parallel_size > 1 {
    match dp_utils::extract_dp_rank(prefill_worker.url()) {
        Ok((base, rank)) => (base.to_string(), Some(rank)),
        ...
    }
} else {
    (prefill_worker.url().to_string(), None)
};
let prefill_url = format!("{}{}", prefill_base_url, path);  // clean URL

// New code
let prefill_url = prefill_worker.endpoint_url(path);
```

There is no test for hostname:port@rank. The underlying URL parser interprets `http://node1:8087@1` as: `node1` → username, `8087` → password, `1` → host → `0.0.0.1`

This is standard HTTP userinfo syntax (user:pass@host), so the parser is correct, the issue appears to be that the final request URL still contains the `@rank` suffix, suggesting endpoint_url() or related URL logic is not stripping `@rank` before handing it to reqwest.

qiching:
## Purpose

Fixes #100 — regression from #98 where hostname:port URLs with
`--intra-node-data-parallel-size > 1` fail because `@rank` suffix
is not stripped before passing to reqwest.

## Root Cause

PDRouter stores DP-expanded URLs (e.g. `http://node1:8087@3`) as
`BasicWorker`. `BasicWorker`'s default `base_url()` returns the raw
URL including `@rank`. #98 replaced manual `extract_dp_rank()` stripping
with `endpoint_url()`, which relies on `base_url()` — but `BasicWorker`
does not override it to strip `@rank`.

When reqwest receives `http://node1:8087@3`, it parses as HTTP userinfo:
- user=`node1`, pass=`8087`, host=`3` → `0.0.0.3`

The old code already handled this correctly via `extract_dp_rank()`.
#98 inadvertently removed that working logic.

## Fix

Replace `prefill_worker.endpoint_url(path)` / `decode_worker.endpoint_url(path)`
with `dp_utils::parse_worker_url(worker.url())` + `format!("{}{}", base, path)`
to explicitly strip `@rank` before constructing request URLs.

## Test Plan

Added unit tests for hostname:port format in dp_utils:
- `test_parse_worker_url_hostname_port_dp`
- `test_parse_worker_url_hostname_port_no_dp`
- `test_reqwest_userinfo_proof` — proves `@rank` causes host corruption
- `test_reqwest_userinfo_proof_ipv6` — proves IPv6 also works correctly
- `test_dp_expand_roundtrip_hostname` — end-to-end hostname:port
- `test_dp_expand_roundtrip_ipv6` — end-to-end IPv6

cargo test -- dp_utils

`15 passed; 0 failed`
Screenshot 2026-03-05 at 4 58 04 PM

**After (this PR):** Requests route correctly to actual worker hostnames.

### DIFF
--- a/src/routers/http/dp_utils.rs
+++ b/src/routers/http/dp_utils.rs
@@ -254,4 +254,123 @@ mod tests {
             assert_eq!(rank, i);
         }
     }
+    // ============================================================
+    // Verify parse_worker_url handles ALL URL formats correctly
+    // These tests cover both the #98 regression (hostname:port)
+    // and the original #92 concern (IPv6)
+    // ============================================================
+
+    #[test]
+    fn test_parse_worker_url_hostname_port_dp() {
+        let (base, rank) = parse_worker_url("http://node1:8087@3");
+        assert_eq!(base, "http://node1:8087");
+        assert_eq!(rank, Some(3));
+        // Verify the URL we'd send to reqwest is clean
+        let request_url = format!("{}/v1/completions", base);
+        assert!(
+            !request_url.contains('@'),
+            "request URL must not contain @rank"
+        );
+        assert_eq!(request_url, "http://node1:8087/v1/completions");
+    }
+
+    #[test]
+    fn test_parse_worker_url_hostname_port_no_dp() {
+        let (base, rank) = parse_worker_url("http://node1:8087");
+        assert_eq!(base, "http://node1:8087");
+        assert_eq!(rank, None);
+    }
+
+    // ============================================================
+    // Critical: prove that reqwest would parse @rank as userinfo
+    // This is the actual bug that #98 introduced
+    // ============================================================
+
+    #[test]
+    fn test_reqwest_userinfo_proof() {
+        // Demonstrate WHY @rank must be stripped before reqwest:
+        // reqwest/url parses "http://host:port@rank" as userinfo
+
+        // BAD: raw URL with @rank — reqwest sees wrong host
+        let bad_url = url::Url::parse("http://node1:8087@3/v1/completions").unwrap();
+        assert_eq!(
+            bad_url.host_str(),
+            Some("0.0.0.3"),
+            "proves @rank corrupts host"
+        );
+        assert_eq!(bad_url.username(), "node1", "host becomes username");
+
+        // GOOD: stripped URL — reqwest sees correct host
+        let (base, rank) = parse_worker_url("http://node1:8087@3");
+        let good_url_str = format!("{}/v1/completions", base);
+        let good_url = url::Url::parse(&good_url_str).unwrap();
+        assert_eq!(
+            good_url.host_str(),
+            Some("node1"),
+            "host is correct after stripping"
+        );
+        assert_eq!(good_url.port(), Some(8087));
+        assert_eq!(rank, Some(3));
+    }
+
+    #[test]
+    fn test_reqwest_userinfo_proof_ipv6() {
+        // Same proof for IPv6 — verify our fix works for both formats
+        let (base, rank) =
+            parse_worker_url("https://[2a03:83e4:5006:0090:5f5a:f8c5:0400:0000]:20009@2");
+        let url_str = format!("{}/v1/completions", base);
+        let parsed = url::Url::parse(&url_str).unwrap();
+        assert_eq!(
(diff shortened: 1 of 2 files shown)


Prowindy (commented):
@qiching thanks for good catch and contributing to fix. Do you agree with the following analysis? comparing https://github.com/vllm-project/router/pull/104 and this PR


Analysis from Claude Code: Comparison of Two Approaches to Fix DP URL Corruption
======================================================

Context: When intra_node_data_parallel_size > 1, worker URLs are expanded
to include @rank (e.g., http://host:8087@1). If this URL is passed directly
to reqwest, it gets misinterpreted as HTTP userinfo (user=host, pass=8087,
host=1), causing connection failures.

Approach A: Use DPAwareWorker at worker creation time
-----------------------------------------------------
Change Router and PDRouter to create DPAwareWorker (instead of BasicWorker)
when dp_size > 1. DPAwareWorker stores the clean base_url separately, so
endpoint_url() and base_url() always return URLs without @rank.

Approach B: Use parse_worker_url at each call site
---------------------------------------------------
Keep using BasicWorker, but at the specific points where the URL is used
for HTTP requests, call dp_utils::parse_worker_url(worker.url()) to strip
the @rank suffix, then manually reconstruct the URL with format!().


Both approaches work correctly
------------------------------
Both approaches produce the same result for the immediate bug, including
IPv6 URLs. parse_worker_url splits on '@', which correctly handles:

  - http://host:8087@1       -> ("http://host:8087", Some(1))
  - https://[2a03:...]:20009@2 -> ("https://[2a03:...]:20009", Some(2))

So functionally, either approach fixes the URL corruption and the IPv6 case.

Summary
-------
Both approaches solve the problem. Approach A makes the fix structural so
that endpoint_url() and base_url() are always safe to use, which reduces
the chance of the bug reappearing if new call sites are added later.
Approach B is more targeted and minimal, fixing only the specific code
paths that currently hit the issue.

Either way, the IPv6 fix works identically since both rely on the same
parse_worker_url / extract_dp_rank logic under the hood.

PR #120, opened 2026-03-17
Migrate /v1/chat/completions from typed struct to JSON pass-through

Friday13th:
## Purpose

  Migrate `/v1/chat/completions` from typed `ChatCompletionRequest` struct to JSON pass-through (`serde_json::Value`).                                                  
                                                                                                                                                                      
  The previous implementation deserialized the request body into a `ChatCompletionRequest` struct before forwarding it to backend workers. This caused requests with
  extra fields (e.g. `reasoning`, custom vLLM parameters) to fail with deserialization errors, even though the router doesn't need to interpret those fields — it just
  forwards them.

  This PR changes `route_chat` across all router types to accept `&serde_json::Value`, allowing arbitrary JSON to pass through unchanged. The `v1_chat_completions`
  handler now parses the body as raw JSON instead of a typed struct.

## Test Plan

  cargo test
  cargo fmt --check
  cargo clippy

  New test files:
  - tests/test_chat_completion_features.rs — 9 tests for Regular Router's route_chat: round-robin routing, streaming/non-streaming, header forwarding, JSON pass-through
   integrity, retry on failure, model-specific routing, and a regression test proving route_chat retries while route_transparent does not.
  - tests/test_vllm_pd_chat.rs — 2 tests for VllmPD Router: error path (503 with no workers) and arbitrary JSON acceptance.

## Test Result
All tests pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,9 @@ target
 
 .tokenizer_cache/*
 
+# VS Code
+.vscode/
+
 # RustRover
 #  JetBrains specific template is maintained in a separate JetBrains.gitignore that can
 #  be found at https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore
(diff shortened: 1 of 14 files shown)


Prowindy (changes requested):
looks good. could you fix buildkite failures?

PR #128, opened 2026-03-26
Add rendezvous hashing (HRW) routing policy

wpc:
## Summary

  - Add a new `rendezvous_hash` load balancing policy using Highest Random Weight (HRW) hashing. For each request, scores every healthy worker with `fbi_hash("{hash_key}:{worker_url}")` and picks the highest score. Fully deterministic, session-sticky, with ~1/n redistribution on worker add/remove.
  - Extract shared hash key extraction logic (headers, body fields, fallback) into a common `hash_key` module, eliminating duplication between `ConsistentHashPolicy` and `RendezvousHashPolicy`.

  ## Why rendezvous over ring-based consistent hashing?

  Simulation comparison (10 workers, 4 DP ranks per worker, 1000 trials):

  | Sessions/rank | Ring CV | Rendezvous CV | CV Improvement | Ring Imbalance | Rendezvous Imbalance | Imb Improvement |
  |---|---|---|---|---|---|---|
  | 16 | 0.2546 | 0.2463 | -3% | 1.604 | 1.582 | -1% |
  | 32 | 0.1873 | 0.1737 | -7% | 1.433 | 1.399 | -2% |
  | 64 | 0.1421 | 0.1229 | -14% | 1.324 | 1.281 | -3% |
  | 128 | 0.1135 | 0.0865 | -24% | 1.250 | 1.195 | -4% |
  | 256 | 0.0946 | 0.0615 | -35% | 1.201 | 1.139 | -5% |
  | 512 | 0.0841 | 0.0432 | **-49%** | 1.168 | 1.097 | **-6%** |

**CV (Coefficient of Variation)** — standard deviation divided by mean of sessions per DP rank. Lower is better. 0.0 = perfectly balanced, 0.10 = 10% typical deviation from mean.
**Imbalance (max/expected)** — ratio of the busiest DP rank's session count to the ideal even split. Lower is better. 1.0 = perfectly balanced, 1.20 = busiest rank has 20% more sessions than expected.

  The ring-based approach has a CV floor (~0.08) caused by uneven arc distribution that more sessions cannot fix. Rendezvous hashing has zero arc bias, so its CV approaches the theoretical minimum (pure binomial noise). It also requires zero tuning — no vnodes, no ring size to configure.

## When to use rendezvous hashing

  - **Session-sticky routing with many concurrent sessions** — the balance advantage grows with session count. At 128+ sessions per DP rank, rendezvous hashing is 24-49% better balanced than ring-based consistent hashing.
  - **DP-parallel serving** — when workers are grouped by DP rank, even small imbalances get amplified because the slowest rank gates the whole group. Rendezvous hashing's lower imbalance ratio (e.g. 1.097 vs 1.168 at 512 sessions/rank) directly reduces tail latency.
  - **Dynamic worker pools** — when workers are frequently added/removed (scaling, rolling deploys), rendezvous hashing redistributes only ~1/n of sessions, same as ring-based, but without needing to rebuild a hash ring or tune vnode counts.
  - **Simplicity** — no configuration needed. Ring-based consistent hashing requires choosing a vnode count (too few = poor balance, too many = memory overhead). Rendezvous hashing has no knobs.

  ## When NOT to use (use consistent hashing instead)

  - **Very large worker pools (hundreds+)** — rendezvous hashing scores every healthy worker per request (O(n) per lookup), whereas ring-based consistent hashing does an O(log n) lookup after ring construction. For small-to-medium pools (≤50 workers) the difference is negligible, but at hundreds of workers the per-request cost may become measurable.

  ## Code Level tests

  - [x] 7 unit tests for `RendezvousHashPolicy`: consistency, distribution, unhealthy workers, no healthy workers, minimal redistribution, header routing
  - [x] 19 unit tests for shared `hash_key` module: field extraction (double/single/unquoted), nested fields, JSON object parsing, header priority, body priority, fallback behavior, edge cases
  - [x] All 56 policy tests pass (`cargo test --lib -- policies`)
  - [x] Simulation example works with both `--mode ring` and `--mode rendezvous`
  - [x] Factory and registry tests updated for new policy

  ## E2E A/B Test
k8s, a large model with gb300 dp8ep decode + prefill disagg setup, 3 hours aiperf bench on both side with multi-turn long input payload. Both side with PR #149 to avoid pinning traffic to rank 0
### Performance
| Metric | consistent_hash | rendezvous_hash | Delta |
|---|---|---|---|
| **Total TPGS** | **3,650.26** | **3,892.01** | **+6.6%** |
| Prefill TPGS | 7,008.28 | 7,472.43 | +6.6% |
| Decode TPGS | 292.24 | 311.60 | +6.6% |
| **TTFT mean** | 24,011 ms | 18,869 ms | **-21.4%** |
| TTFT p50 | 759 ms | 696 ms | -8.3% |
| **TTFT p95** | 59,029 ms | 42,020 ms | **-28.8%** |
| TTFT p99 | 417,503 ms | 364,591 ms | -12.7% |
| ITL mean | 22.93 ms | 22.45 ms | -2.1% |
| ITL p50 | 23.14 ms | 22.19 ms | -4.1% |
| ITL p99 | 45.28 ms | 41.71 ms | -7.9% |
| Benchmark duration | 10,826 s | 10,828 s | same |
| Decode errors | 281 | 134 | -52% |
### Host Level Load Balance 
| Metric | consistent_hash | rendezvous_hash |
|---|---|---|
| Total routing decisions | 38,866 | 42,332 |
| Workers | 16 | 16 |
| Coefficient of Variation | 18.1% | 17.2% |
| Max/Min ratio | 1.92 | 2.16 |
| Prefill host balance | 9,761 vs 9,671 | 11,222 vs 9,943 |
| Decode host balance | 10,282 vs 9,148 | 10,451 vs 10,714 |
### Engine Level Load Balance
Per engine CV compare (on decode node) shows hrw is more balance most of the time
Screenshot 2026-04-16 at 2 48 29 AM

### DIFF
--- a/examples/simulate_consistent_hash.rs
+++ b/examples/simulate_consistent_hash.rs
@@ -330,10 +330,7 @@ fn analyze_ring_distribution(
 
     for (w, &vnode_count) in vnodes_per_worker.iter().enumerate() {
         let pct = 100.0 * vnode_count as f64 / total_vnodes as f64;
-        println!(
-            "  decode-worker{}: {} vnodes ({:.1}%)",
-            w, vnode_count, pct
-        );
+        println!("  decode-worker{}: {} vnodes ({:.1}%)", w, vnode_count, pct);
     }
 
     let hashes: Vec<u64> = ring.keys().copied().collect();
@@ -361,12 +358,7 @@ fn analyze_ring_distribution(
 
     println!("\nHash ring arc ownership (fraction of keyspace):");
     for (w, &arc) in arc_per_worker.iter().enumerate() {
-        println!(
-            "  decode-worker{}: {:.4} ({:.2}%)",
-            w,
-            arc,
-            arc * 100.0
-        );
+        println!("  decode-worker{}: {:.4} ({:.2}%)", w, arc, arc * 100.0);
     }
     let arc_min = arc_per_worker.iter().cloned().fold(f64::INFINITY, f64::min);
     let arc_max = arc_per_worker
--- a/examples/simulate_consistent_hash.rs
+++ b/examples/simulate_consistent_hash.rs
@@ -197,7 +197,10 @@ fn run_single_trial(args: &Args, verbose: bool) -> TrialStats {
         println!("Expected per worker:  {:.1}", mean);
         println!("Min:                  {}", min_count);
         println!("Max:                  {}", max_count);
-        println!("Imbalance ratio:      {:.4} (max/expected)", worker_imbalance);
+        println!(
+            "Imbalance ratio:      {:.4} (max/expected)",
+            worker_imbalance
+        );
         println!("Std deviation:        {:.2}", std_dev);
         println!("Coeff of variation:   {:.4} (lower is better)", worker_cv);
 
@@ -366,10 +369,7 @@ fn analyze_ring_distribution(
             arc_per_worker[w] * 100.0
         );
     }
-    let arc_min = arc_per_worker
-        .iter()
-        .cloned()
-        .fold(f64::INFINITY, f64::min);
+    let arc_min = arc_per_worker.iter().cloned().fold(f64::INFINITY, f64::min);
     let arc_max = arc_per_worker
         .iter()
         .cloned()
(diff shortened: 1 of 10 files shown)


Prowindy:
Follow-up (based on offline discussion with Pengchao): It should work as a direct replacement in most cases. We can remove the consistent hashing path once we've validated there are no remaining dependencies.

PR #154, opened 2026-04-28
ci: use cuda 13

BugenZhao:
The vLLM 0.20.0 release has switched to CUDA 13 for the default wheels. Let's also bump the CUDA version used in CI from 12.9 to 13.0 to fix failures like https://buildkite.com/vllm/router/builds/477#019dd2fa-aa7e-4d77-a3c6-101a846b1707/L2205.


liuzijing2014 (approved):
LGTM!

PR #155, opened 2026-04-28
refactor: remove old http-based pd impl

BugenZhao:
The old `prefill_decode` mode is not actually applicable to vLLM PD deployment. Instead, it currently serves only as the base implementation of the effective `vllm_prefill_decode` mode, and a large portion of it is dead code.

This PR cleans it up.


liuzijing2014 (approved):
LGTM!

PR #165, opened 2026-05-03
fix: attribute transparent-proxy traffic to run_id

JannikSt:
Catch-all transparent proxy was bypassing per-run usage counters.

- `/v1/generate` (renderer rollouts) is not an explicit route on the router, so requests fall through to `transparent_proxy_handler` → `route_transparent`, which never called `record_run_request` / `extract_and_record_usage`.
- Result on Johannes' run `o2eh6s2nqx41l4pq5qfft8jk`: 0 entries in `vllm_router_run_prompt_tokens_total{run_id=...}` despite millions of tokens generated and 200 OK responses on `/v1/generate` at the inference pods.

Threads `run_id` through `RouterTrait::route_transparent` and reuses the existing `record_run_request` + `extract_and_record_usage_buffered` helpers. Streaming pass-through preserved when no `run_id` is present.


JannikSt:
Wrong upstream — re-opening on PrimeIntellect-ai/router fork.

PR #171, opened 2026-05-05
feat: export queue metrics and make queue size/timeout configurable

pierre13hk:
## Purpose
- Being able to configure queue size and timeout is useful for tailoring the router to different traffic types. For example a larger queue size is useful for bursty traffic.
- Exporting the queue metrics to prometheus allows observability into the queue state and scaling based on queue size with tools like Keda for example.

## Test Plan
- Ran the project locally with dummy backends: cargo run -- --worker-urls http://worker1:8000 http://worker2:8000 --queue-size 500 --queue-timeout-secs 300
- Added unit tests similar to those already present for `RouterMetrics`, but am unsure if this is enough?

## Test Result
- Starting the project locally with the above command works
- Tests pass locally
---
<details>

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -251,6 +251,14 @@ struct CliArgs {
     #[arg(long, default_value_t = 32768)]
     max_concurrent_requests: usize,
 
+    /// Maximum capacity of the requests queue
+    #[arg(long, default_value_t = 100)]
+    queue_size: usize,
+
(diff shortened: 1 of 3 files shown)


pierre13hk:
Hello, while i no longer need the contents of this PR, i'd be happy to correct if need be or close if it's not a wanted feature.

PR #180, opened 2026-05-31
feat(router): register /inference/v1/generate as a typed route

### LINKED ISSUE #179: cache_aware degenerates to a single worker on transparent-proxy endpoints (route_transparent never tracks worker load)
## Summary

Requests served through the **transparent proxy** path (`route_transparent`, `src/routers/http/router.rs`) are routed by the configured load-balancing policy but the path **never calls `increment_load` / `decrement_load`**. For the default `cache_aware` policy this makes every worker permanently appear at load 0, so its shortest-queue rebalancing can never engage and **all requests that share a prefix collapse onto a single worker** while the others sit idle.

This affects any endpoint that is not a first-class typed route and therefore falls through to the transparent proxy — e.g. `/inference/v1/generate` (used by SLIME/vime-style RL rollout). In an 8-engine deployment we observed one engine serving 100% of requests (480 concurrent) while 7 were idle, and the lone engine eventually OOM'd its KV cache.

## Root cause

`route_typed_request` tracks load for `cache_aware` (around `src/routers/http/router.rs:574`):

```rust
let load_incremented = if policy.name() == "cache_aware" {
    worker.increment_load();
    RouterMetrics::set_running_requests(worker.url(), worker.load());
    true
} else { false };
// ... decremented on completion / error / stream end
```

`route_transparent` selects a worker via the same policy but has **no equivalent** increment/decrement. So in `cache_aware` (`src/policies/cache_aware.rs`):

```rust
let is_imbalanced = max_load.saturating_sub(min_load) > self.config.balance_abs_threshold
    && (max_load as f32) > (min_load as f32 * self.config.balance_rel_threshold);
```

`max_load == min_load == 0` forever → `is_imbalanced` is always false → the policy stays in pure prefix-affinity mode → one worker absorbs everything.

(Note: the transparent path also routes on `serde_json::to_string(&body)` as the cache key, which is a separate quality issue, but the missing load tracking is what causes the single-worker collapse.)

## Impact

- Transparent-proxy endpoints (e.g. `/inference/v1/generate`) with the default `cache_aware` policy use 1/N of the engines and can OOM the lone worker.
- Silent: no error, just degraded utilization / a crash under load.

## Fix

Track worker load in `route_transparent` for `cache_aware`, mirroring `route_typed_request`: increment after selection, decrement when the proxied response stream completes or the upstream request fails. PR incoming.

Found while diagnosing single-engine pileup on `/inference/v1/generate`; AI assistance (Claude Code) was used in the investigation.

aoshen02:
Fixes #179.

## Problem

vLLM's disaggregated engine serves generation at `/inference/v1/generate` (the `/inference/v1/` prefix is vendor-private). The router has no first-class typed route for it, so it falls through to the **transparent proxy**. On that path `cache_aware` never tracks worker load, so it degrades to single-worker prefix affinity — one engine serves every shared-prefix request (and can OOM its KV cache) while the others idle.

## Fix

Register `/inference/v1/generate` as a **typed route**, so it reuses `route_typed_request`'s `cache_aware` routing + worker load tracking + retry + circuit breaker + metrics — identical to `/generate`.

The typed path forwards by re-serializing the request struct, so a new **lossless** `InferenceGenerateRequest` models only what the router reads — `token_ids` (the cache_aware routing key), `stream`, and `model` (for `get_model`, consistent with the other typed endpoints) — and passes everything else through with `#[serde(flatten)]`. The nested `sampling_params` is forwarded untouched, so `seed` / `max_tokens` / `logprobs` reach the engine unchanged (the router never interprets sampling params; mirroring vLLM's ~100-field `SamplingParams` would be pure drift risk). `GenerateRequest` and `/generate` are untouched.

## Test

Unit tests for `InferenceGenerateRequest`: routing key = `token_ids` joined as decimal string; **lossless round-trip** (`sampling_params.{max_tokens,seed,logprobs,temperature}` + `model` survive deserialize → serialize); same `token_ids` / different `seed` → identical routing key; `is_stream` / `get_model`.

> Supersedes the earlier transparent-proxy load-tracking approach previously on this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/protocols/spec.rs
+++ b/src/protocols/spec.rs
@@ -1971,6 +1971,42 @@ impl GenerationRequest for GenerateRequest {
 }
 
 // ==================================================================
+// =   VLLM SPEC - DISAGG /inference/v1/generate API              =
+// ==================================================================
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct InferenceGenerateRequest {
+    pub token_ids: Vec<i32>,
+
+    #[serde(default)]
+    pub stream: bool,
+
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub model: Option<String>,
+
+    #[serde(flatten)]
+    pub other: serde_json::Map<String, serde_json::Value>,
+}
+
+impl GenerationRequest for InferenceGenerateRequest {
+    fn is_stream(&self) -> bool {
+        self.stream
+    }
+
+    fn get_model(&self) -> Option<&str> {
+        self.model.as_deref()
+    }
+
+    fn extract_text_for_routing(&self) -> String {
+        self.token_ids
+            .iter()
+            .map(|id| id.to_string())
+            .collect::<Vec<String>>()
+            .join(" ")
+    }
+}
+
+// ==================================================================
 // =            VLLM SPEC - RERANK API                            =
 // ==================================================================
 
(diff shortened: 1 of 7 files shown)


aoshen02:
## Validation

Built this branch and ran it against **4 live vLLM engines**. The test sends **256 requests** to `/inference/v1/generate`, structured as **32 distinct prompts × 8 requests that share the same prompt prefix** (a best-of-n / multi-sample pattern), and sweeps all three policies.

**End to End routing (4 workers, 256 requests):**

| policy | per-worker load | engine prefix-cache hit |
|---|---|---|
| `round_robin` | 64 / 64 / 64 / 64 | ~60% |
| `cache_aware` | 64 / 64 / 64 / 64 | **~86%** |
| `consistent_hash` | 61 / 81 / 57 / 57 | ~60% |

### Where the 86% comes from

Each of the 32 prompts is sent as a group of 8 requests sharing the full prompt prefix. `cache_aware` routes all 8 of a group to the **same** worker, so that worker computes the prefix **once** and reuses it for the other 7 → ceiling = (8−1)/8 = **87.5%**, and we measure ~86% (essentially the ceiling). `round_robin` scatters each group across workers, so the prefix gets recomputed on several engines → only ~60%.

The per-request decision trace confirms the mechanism directly:

```
match_rate=0.00 matched=0/873   branch=MIN_LOAD  chosen=:15006 tenant=:15006   # new prompt, tree empty → least-loaded worker, seeds the tree
match_rate=1.00 matched=873/873 branch=CACHE_HIT chosen=:15006 tenant=:15006   # later request of same group → full match → sticks to :15006
```

223 of 256 requests matched at rate 1.0 — and they group into 31 distinct prefix lengths, each appearing ~7×, i.e. exactly the "7 reuses per 8-request group".

### Why this shows cache_aware actually doing its job (not degenerate, and not just consistent_hash)

The fix's value is that cache_aware now gets that reuse **without hot-spotting**. On the old transparent-proxy path it had no load counter, so affinity meant one worker served every shared-prefix request while the others idled (and risked KV-cache OOM). On the typed route it achieves **~86% reuse AND an even 64/64/64/64 load at the same time**.

What makes that possible — and distinguishes it from a pure hash policy — is the load-vs-cache arbitration in the trace: when a request's best-prefix-match worker (`tenant`) is only weakly matched (below `cache_threshold = 0.3`), it is redirected to the **least-loaded** worker instead of piling onto the prefix owner. In this run, of the requests that took the min-load branch, **17 of 22 had `tenant != chosen_worker`** — a load override that `consistent_hash` cannot do (it routes purely by key hash, which is why it skews to 81 vs 57 with no rebalancing).

BugenZhao (approved):

tkorochka:
doh

PR #181, opened 2026-06-02
fix(pd): route MoRIIO KV-notify to the correct decode DP rank in 2P2D

raviguptaamd:
Rebased onto current `main` (`1d10e71`) and reduced to a single commit. **This PR was previously 2 commits and conflicted; it is now 1 commit, no conflicts.**

### What changed about this PR

The round-robin DP-rank commit that used to ride along here (`11841c0d`) is **already upstream** — `prefill_dp_round_robin` is in `vllm_pd_router.rs` on `main` today. Keeping it was the main source of the conflicts, so it's dropped. What remains is the one fix that is still missing upstream.

### The fix

In 2P2D disaggregated serving over the MoRIIO connector, the decode side's KV-notify is addressed using the **intra-node** DP size, so with more than one decode pod the notify targets the wrong DP rank. The transfer is then never acknowledged and the request wedges until the deferred write expires:

```
remote blocks never arrived
Reaped deferred sends
```

This adds `moriio_dp_size` and routes notify through `effective_dp_size()`, falling back to the intra-node value when the flag isn't set — so single-pod behaviour is byte-for-byte unchanged. `remote_dp_rank_override` carries the resolved rank across the pod boundary.

`--moriio-dp-size` is **opt-in**. Without it, `effective_dp_size()` returns the intra-node size and nothing about existing deployments changes.

### Validation

Built into a GLM-5.1-FP8 disaggregated image and run on MI300X (8 nodes, 2P/2D EP16, MoRI-EP wideEP):

- **NIAH 3 seeds:** 2k 9.7/10, 8k 9.7/10
- **Perf 8192/1024 con32:** TPOT 60.0 ms median, 440 output tok/s, 0 failed requests
- No `remote blocks never arrived`, no `Reaped deferred sends` — the wedge this fixes does not reproduce

Also perf-neutral against the pre-rebase router: 60.0 ms / 440 tok/s here vs 59.5 ms / 444 tok/s on the old branch, i.e. within noise. The router compiles clean (`vllm_router_rs`, release).

Consumed by ROCm/MAD#206, which currently has to pin a fork of this repo until this merges.


raviguptaamd:
Force-pushed this branch to rebase it — worth flagging what changed, since the diff will look different to anyone who reviewed it earlier.

**Before:** 2 commits, based on June `main`, `mergeable=false` (conflicted).
**Now:** 1 commit on top of `1d10e71`, no conflicts.

The dropped commit is the round-robin DP-rank fix (`11841c0d`). It's **already upstream** — `prefill_dp_round_robin` is in `vllm_pd_router.rs` on `main`, so carrying it here was both redundant and the main source of the conflicts. Resolving the rebase meant keeping upstream's counter and driving it from `effective_dp_size()` rather than reintroducing my duplicate.

The previous head `6409ac1` is preserved at tag `pr181-pre-rebase-6409ac1` on the fork if anyone wants to diff against it.

Nothing about the remaining fix changed — only its base.

raviguptaamd:
Superseded by #223 — closing this.

Same fix, but this PR was opened 2026-06-02 against a much older `main`, and the two review comments on it predate today's rebase, so the diff they refer to no longer exists. Rather than ask anyone to re-read a PR whose contents changed underneath them, #223 is the same commit (`82dc9811`) as a clean single-commit PR against current `main`.

The round-robin commit that used to ride along here (`11841c0d`) is already upstream as `prefill_dp_round_robin`, so it's dropped — only the 2P2D KV-notify fix remains.

Old head `6409ac1` is preserved at tag `pr181-pre-rebase-6409ac1` on the fork if anyone wants to diff against it.

PR #204, opened 2026-08-11
test: remove integration test network races

andyluo7:
## What changed

- keep the mock worker's ephemeral TCP listener bound and pass it directly to Axum
- run the OpenAI health test against an auth-required local mock server

## Root cause

`MockWorker::start` previously bound port `0`, recorded the assigned port,
dropped the listener, and rebound the port in a spawned task. Parallel tests
could claim the port during that gap, producing `Address already in use` and a
downstream HTTP 500.

The OpenAI health test also called `api.openai.com` directly. A transient
network failure produced HTTP 503 even though the Router code was unchanged.

## Impact

Rust integration tests no longer depend on an external service or use a
time-of-check/time-of-use port reservation. The local health fixture requires
auth, preserving coverage that an unauthenticated 401 response still proves
the upstream endpoint is reachable.

## Validation

Run in `rust:1.95.0-bullseye`, matching Buildkite:

- `cargo fmt --check`
- each formerly failing test passed 20 consecutive repetitions
- authenticated health test passed 20 consecutive repetitions after review update
- `test_openai_routing`: 12 passed
- `test_dp_routing`: 16 passed

### DIFF
--- a/tests/common/mock_worker.rs
+++ b/tests/common/mock_worker.rs
@@ -61,18 +61,14 @@ impl MockWorker {
     /// Start the mock worker server
     pub async fn start(&mut self) -> Result<String, Box<dyn std::error::Error>> {
         let config = self.config.clone();
-        let port = config.read().await.port;
+        let requested_port = config.read().await.port;
 
-        // If port is 0, find an available port
-        let port = if port == 0 {
-            let listener = std::net::TcpListener::bind("127.0.0.1:0")?;
-            let port = listener.local_addr()?.port();
-            drop(listener);
-            config.write().await.port = port;
-            port
-        } else {
-            port
-        };
+        // Bind once and pass the listener directly to axum. Reserving a port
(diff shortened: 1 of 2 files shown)


andyluo7:
CI triage update:

- Buildkite #644 Rust Integration Tests passed on the signed-off head with zero retries.
- The remaining Buildkite wait is the existing generic 4-GPU P/D lane; the ROCm lanes are intentionally ignored on this non-ROCm PR build.
- The Trivy failure is baseline-wide: the same msgpack/setuptools findings also fail current `main` (Actions run 31441755053).
- Before push, both formerly flaky tests passed 20 consecutive repetitions each in the same `rust:1.95.0-bullseye` image used by Buildkite; the full `test_openai_routing` (12/12) and `test_dp_routing` (16/16) suites also passed.

BugenZhao (approved):
LGTM. Thanks!

PR #205, opened 2026-08-11
ci: serialize ROCm cluster jobs

andyluo7:
## What changed

- serialize XGMI and two-node RDMA jobs through one organization-wide
  `vllm/router-rocm-mi300x-pair` concurrency group
- automatically retry only Buildkite's lost-agent exit status (`-1`)
- retain inspected manual retries for GPU and RDMA failures

## Why

The queue currently has one coordinator agent, but queue size alone is not a
durable resource lock. A future agent or another pipeline could otherwise own
the same coordinator/worker pair concurrently.

Wildcard automatic retries also allow a functional GPU or RDMA regression to
finish green on a second attempt. Limiting automatic retry to a lost agent keeps
infrastructure recovery while preserving real failures for triage.

## Validation

- generic YAML parsing passed
- the live Buildkite agent accepted the complete pipeline with
  `pipeline upload --dry-run --no-interpolation`
- rendered output contained `concurrency: 1`, the shared concurrency group, and
  `exit_status: -1` on both ROCm steps

### DIFF
--- a/.buildkite/README.md
+++ b/.buildkite/README.md
@@ -83,6 +83,14 @@ ROCm-only scheduled builds should set both `NIGHTLY=1` and `ROCM_ONLY=1`.
 `ROCM_ONLY=1` skips the unrelated NVIDIA four-GPU P/D step while retaining the
 CPU build that supplies the router artifact to both ROCm matrices.
 
+Both ROCm steps share the organization-wide concurrency group
+`vllm/router-rocm-mi300x-pair` with a limit of one. This prevents jobs from
+different builds or pipelines from owning the coordinator/worker pair at the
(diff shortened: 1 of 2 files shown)


andyluo7:
Buildkite #645 failed only in the known `test_openai_router_health` path (`503` instead of `200`). That race is fixed by sibling PR #204, whose Rust Integration Tests passed on Buildkite #644. This scheduling PR is intentionally kept independent; after #204 merges, it should be refreshed onto `main` and Buildkite rerun. The Trivy failure is also present on current `main` (Actions run 31441755053), so it is unrelated to this YAML-only diff.

PR #207, opened 2026-08-12
feat: concurrent P/D dispatch for NIXL (decode prepares while prefill runs)

dczhu:
Companion of the vLLM PR https://github.com/vllm-project/vllm/pull/51919 (requires its
`/v1/pd_kv_ready` + prepare-parking support).

### Motivation

The two-stage vLLM PD flow is strictly serial: the decode worker receives the
request only after the prefill has responded, so decode-side HTTP parsing,
tokenization and scheduler admission (~32ms p50 on our 3P1D setup) sit on the
TTFT critical path.

### Changes

With `VLLM_PD_CONCURRENT_DISPATCH=1` (opt-in; NIXL connector only), the router
dispatches both legs at once:

- The decode leg is spawned first with `kv_transfer_params=
  {"do_remote_decode": false, "do_remote_prefill": true, "remote_ready": false}`;
  the decode instance tokenizes, enqueues and parks the request while the
  prefill computes.
- When the prefill responds, the router forwards its `kv_transfer_params`
  (including seeded `first_token_ids`, when enabled) to the decode worker via
  `POST /v1/pd_kv_ready`, retried 3 times.
- Failures (prefill error, missing `kv_transfer_params`, undeliverable ready
  POST) abort the spawned decode leg and return an error; an abort-on-drop
  guard covers client disconnects; a notification lost after a 2xx is covered
  by the decode-side park timeout (local prefill fallback).
- Response handling (logprobs merge / streaming passthrough) and PD metrics
  mirror the serial path. Default off; the serial flow is unchanged.

### Testing

- 3P1D deployment with the vLLM PR; flag on/off A/B runs (1000 requests) and
  fault injection (dropped ready POST, prefill failure, client disconnect).
- Result: decode front-end fully overlapped with prefill; handover
  63ms → 52/36ms p50, mean/median TTFT −23% vs. stock at default settings.

### Changes in the 8/26 update (addressing review comments) (re-tested)
- **Env var → config option.** Removed the `VLLM_PD_CONCURRENT_DISPATCH` env var
  (previously read on every request). Concurrent dispatch is now enabled via a proper
  router option: `--pd-concurrent-dispatch` CLI flag / `RouterConfig.pd_concurrent_dispatch`
  (default: off), parsed once at startup.
- **Gated on NIXL pull mode via `transfer_mode`.** The router now learns the connector's
  `transfer_mode` from the first prefill response's `kv_transfer_params` (exposed by
  vllm-project/vllm#50620) and caches it. Concurrent dispatch only activates once
  `"pull"` mode is confirmed; until then (and always for `"push"`), requests take the
  existing serial path. This keeps this PR strictly scoped to pull mode and orthogonal
  to the push-mode concurrent dispatch in #187 — push deployments are never affected.
- **Fixed `/v1/pd_kv_ready` request headers.** The ready notification now carries the
  same headers as the decode leg: `Authorization`, `X-Request-Id`, `Content-Type`, and
  the data-parallel-rank header (`endpoint_url()` strips the `@rank` suffix, so the DP
  rank must travel in the header). Previously the ready POST could be rejected by
  auth-enabled workers or land on the wrong DP rank.

No changes to the vLLM-side protocol; the fallback behavior (decode-side park timeout
covering a lost ready notification) is unchanged.

### DIFF
--- a/src/routers/http/vllm_pd_router.rs
+++ b/src/routers/http/vllm_pd_router.rs
@@ -21,7 +21,7 @@ use axum::{
 use serde_json::{json, Value};
 use std::collections::HashMap;
 use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::Arc;
+use std::sync::{Arc, OnceLock};
 use std::time::Instant;
 use tokio::sync::Mutex;
 use tracing::{debug, error, info, warn};
@@ -61,6 +61,12 @@ pub struct VllmPDRouter {
     kv_connector: KvConnector,
     /// Mooncake bootstrap info: prefill base_url -> MooncakePrefillInfo
     mooncake_prefill_info: Arc<Mutex<HashMap<String, MooncakePrefillInfo>>>,
+    /// Enable concurrent P/D dispatch (NIXL pull mode only)
+    pd_concurrent_dispatch: bool,
+    /// NIXL transfer mode ("pull"/"push") learned from the first prefill
+    /// response's kv_transfer_params; concurrent dispatch stays off until
+    /// pull mode is confirmed.
+    nixl_transfer_mode: Arc<OnceLock<String>>,
 }
 
 /// Transfer ID prefix used by MoRI-IO to correlate prefill and decode legs.
@@ -331,7 +337,11 @@ impl VllmPDRouter {
                     Some(params)
                 }
             }
-            KvConnector::Nixl => Some(prefill_response_json?.get("kv_transfer_params")?.clone()),
+            KvConnector::Nixl => {
+                let params = prefill_response_json?.get("kv_transfer_params")?.clone();
+                self.note_nixl_transfer_mode(&params);
+                Some(params)
+            }
         }
     }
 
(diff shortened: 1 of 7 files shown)


iyastreb:
Is it the same as https://github.com/vllm-project/router/pull/187 ?

src/routers/http/vllm_pd_router.rs
```diff
@@ -1152,6 +1152,359 @@ impl VllmPDRouter {
         .await
     }
 
+    /// Concurrent P/D dispatch kill switch (opt-in, NIXL only).
+    fn concurrent_dispatch_enabled() -> bool {
+        std::env::var("VLLM_PD_CONCURRENT_DISPATCH")
```
iyastreb, line 1157: I think enablement should be a config option, not env var
Maybe by default we should always use concurrent mode

src/routers/http/vllm_pd_router.rs
```diff
@@ -1152,6 +1152,359 @@ impl VllmPDRouter {
         .await
     }
 
+    /// Concurrent P/D dispatch kill switch (opt-in, NIXL only).
+    fn concurrent_dispatch_enabled() -> bool {
```
iyastreb, line 1156: is env var evaluated on each call?

src/routers/http/vllm_pd_router.rs
```diff
@@ -1165,6 +1518,17 @@ impl VllmPDRouter {
         path: &str,
         headers: Option<&HeaderMap>,
     ) -> Result<Response, PDRouterError> {
+        if matches!(self.kv_connector, KvConnector::Nixl) && Self::concurrent_dispatch_enabled() {
```
iyastreb, line 1521: we can gate on NIXL transfer mode, not just the connector

src/routers/http/vllm_pd_router.rs
```diff
@@ -1152,6 +1152,359 @@ impl VllmPDRouter {
...
+        // Prefill request: same shape as the serial path.
+        let mut prefill_request = Self::prepare_prefill_request(original_request.clone(), path);
+        let decode_base_url = decode_worker.base_url().to_string();
+        prefill_request["kv_transfer_params"] = self
+            .build_prefill_kv_transfer_params(None, Some(&decode_base_url), decode_worker.dp_rank())
+            .map_err(|reason| PDRouterError::InvalidConfiguration { reason })?;
+
+        // Decode prepare request: marks the request for scheduler parking.
+        let mut decode_request = original_request.clone();
+        decode_request["kv_transfer_params"] = json!({
```
iyastreb, line 1198: Do we really need anything to be configured on prefill/decode to make this working?

PR #216, opened 2026-08-16
[Bugfix]: Fix health checker to preserve worker load during checks

### LINKED ISSUE #197: [Bug]: Periodic health checks reset active worker load counters and cause cache-aware hot spots
### Your current environment

<details>
<summary>Environment Information</summary>

```text
vLLM Router version: 0.1.14
Rust version: 1.97.1
Python version: 3.10.12
vLLM backend version: 0.26.0
```

</details>

### Component

Load Balancing Policy (cache-aware, consistent-hash, power-of-two, etc.)

### Router Configuration

```
vllm-router \
    --worker-urls \
        http://worker-01:23333 \
        http://worker-02:23333 \
        http://worker-15:23333 \
    --policy cache_aware \
    --intra-node-data-parallel-size 1 \
    --cache-threshold 0.5 \
    --balance-abs-threshold 4 \
    --balance-rel-threshold 1.1 \
    --prometheus-port 7291
```

### Summary

The production health checker periodically resets the local load counter of every registered worker to zero, even when requests are still in flight.

The cache-aware prefix tree is not reset at the same time. As a result, the cache-aware policy sees all workers as balanced and continues routing requests according to historical prefix affinity, even when a worker is already saturated.

This can create a persistent positive feedback loop:

 worker receives more requests
    -> more prefixes are associated with that worker
    -> periodic load reset makes the worker appear idle
    -> prefix-affine requests continue going to that worker
    -> the worker becomes increasingly overloaded

### Root cause

The server starts WorkerRegistry::start_health_checker():

  - server.rs (https://github.com/vllm-project/router/blob/d2ba586533de99ad8f53a55ccd52f3c9cc2963ad/src/server.rs#L953-L960)

That health checker resets every registered worker after every ten health-check cycles, without checking whether requests are active:

  - worker_registry.rs (https://github.com/vllm-project/router/blob/d2ba586533de99ad8f53a55ccd52f3c9cc2963ad/src/core/worker_registry.rs#L355-L398)

With the default 60-second health-check interval, this happens approximately every ten minutes.

BasicWorker::reset_load() performs an unconditional atomic store of zero:

  - worker.rs (https://github.com/vllm-project/router/blob/d2ba586533de99ad8f53a55ccd52f3c9cc2963ad/src/core/worker.rs#L445-L463)

Meanwhile, the cache-aware policy decides whether to use shortest-queue routing from these local counters. If the counters appear balanced, a prefix match above cache_threshold routes directly to the prefix owner:

  - cache_aware.rs (https://github.com/vllm-project/router/blob/d2ba586533de99ad8f53a55ccd52f3c9cc2963ad/src/policies/cache_aware.rs#L245-L325)

The prefix tree is not cleared or reconciled by reset_load().

There are also two health-checker implementations in the repository. The standalone implementation has a conditional low-load reset, while the registry implementation used by the server resets all workers unconditionally. This appears to be an accidental behavioral inconsistency.

### Production observation

In a deployment with 15 healthy vLLM workers, a synchronized snapshot showed:

Router-reported load for the hot worker: 3

Backend metrics for the same worker:
  vllm:num_requests_running:             15
  vllm:num_requests_waiting:              7
  waiting due to capacity:                6
  KV-cache usage:                       ~98%

All workers appeared to the router to have a load between 2 and 3, so cache-aware routing considered the cluster balanced.
Therefore, almost all requests remained on the cache-affinity path instead of triggering shortest-queue balancing.
The backend workers had equivalent KV-cache configurations, so this was not caused by different worker capacity.

### Expected behavior

Health checks must not discard active request accounting.

A worker’s router-side load should remain equal to the number of active proxied requests until those requests complete, fail, or are cancelled.

Cache-aware routing should enter shortest-queue mode when a prefix-affine worker is overloaded.

### Minimal safe fix

Remove the unconditional load-reset block from WorkerRegistry::start_health_checker().

Health checking and load accounting are separate concerns. A worker health check should not mutate the active-request counter.

This preserves:

- Runtime worker health detection.
- Existing cache-aware prefix affinity.
- Correct increment/decrement request accounting.

### Additional context

_No response_

wuhang2014:
## Purpose

Fixes #197: the runtime health checker unconditionally reset every worker's load counter to zero every 10 health-check cycles, even while requests were in flight. The cache-aware prefix tree was never reset, so the policy saw all workers as balanced and kept routing by prefix affinity to already-saturated workers, creating a positive feedback loop of hot spots.

Changes:

- Remove the periodic load reset from `WorkerRegistry::start_health_checker()`; health checks now run in parallel and never mutate load accounting. Load counters are neither reset periodically nor reset on health transitions: a failed `/health` probe does not cancel in-flight requests, so the load a worker reports at recovery time can still be real work.
- Delete the unused standalone `start_health_checker()` with its conditional low-load reset (zero callers).
- Redesign `WorkerLoadGuard` as an owning RAII guard with `share()`/`release()` and adopt it in the OpenAI router proxy path. This also fixes a pre-existing double-decrement on retryable responses and guarantees paired increment/decrement on every path (success, failure, retry, stream abort, client disconnect).
- Adopt `WorkerLoadGuard` for both phases of the P/D two-stage flow (`process_vllm_two_stage_request`), fixing a permanent load leak when a P/D request task is cancelled (e.g. client disconnect) between the manual increment and a decrement branch. Cancellation now releases the prefill/decode load via guard drop; covered by cancellation tests for both phases.
- Clamp decrement at zero with a warning and wire the `vllm_router_worker_load` gauge on increment/decrement/reset.

## Changes since first review

- Removed the unhealthy-to-healthy `reset_load()` call that the first review round flagged: it could erase real in-flight load (a failed `/health` probe does not cancel requests) and raced with new routing. The health checker now never mutates load accounting; `reset_load` is documented as an administrative operation that is only safe when the caller has established that no requests are in flight.
- Strengthened the #197 regression test: it now runs under paused Tokio time with a no-I/O health check and advances through 12 health-check cycles, so it fails against the pre-fix 10-cycle reset (verified by mutation probe: `left: 0, right: 5`).
- Added P/D cancellation tests for both phases (see Test Plan); verified they fail against the previous manual accounting (`left: 1, right: 0`) and pass with the guard-based fix.

## Test Plan

```bash
cargo fmt --check
cargo clippy --all-targets --all-features -- -D warnings
cargo test --lib --bins
cargo test --test '*'
pre-commit run --files src/core/mod.rs src/core/worker.rs src/core/worker_registry.rs \
  src/routers/http/router.rs src/routers/http/vllm_pd_router.rs Cargo.toml
```

New tests:

- `src/core/worker.rs`: `WorkerLoadGuard` unit tests (drop decrements, release idempotence, shared handle decrements once, multi-worker, decrement-at-zero clamps).
- `src/core/worker_registry.rs`: `test_health_checker_preserves_inflight_load` (regression test for #197; runs under paused Tokio time with a no-I/O health check and advances through 12 cycles, so it fails against the pre-fix 10-cycle reset) and `test_health_checker_preserves_load_across_recovery`.
- `src/routers/http/vllm_pd_router.rs`: `test_pd_request_cancellation_releases_prefill_load` and `test_pd_request_cancellation_releases_decode_load` — run the real two-stage flow against mock prefill/decode servers, abort the request while the upstream response is pending, and assert load returns to zero (verified to fail against the previous manual accounting).

## Test Result

All Rust unit/integration tests pass: `cargo test --lib --bins` (494 passed), `cargo test --test '*'` (all suites passed), `cargo fmt --check` and CI-grade clippy clean, all pre-commit hooks pass.

End-to-end verification with 3 mock workers (requests held in flight, 1s health-check interval) comparing router-reported load vs actual in-flight requests:

| Event | Before | After |
|---|---|---|
| 10 requests in flight, 10 health-check cycles elapsed | `Resetting worker loads (cycle 10)`; router view `max_load=0, min_load=0` | No reset lines; `max_load=4, min_load=3` (≈ actual in-flight) |
| Probes while 13-15 requests in flight | Router view `max_load=0..1`, `is_imbalanced=false` -> prefix-affinity routing continues | `max_load=4..5` tracking actual in-flight; balancing decisions see true load |

Mutation probes: restoring the old 10-cycle reset makes the regression test fail (`left: 0, right: 5`); restoring the manual P/D accounting makes both cancellation tests fail (`left: 1, right: 0`).

---

<details>

</details>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,6 +89,7 @@ portpicker = "0.1"
 tempfile = "3.8"
 lazy_static = "1.4"
 opentelemetry_sdk = { version = "0.27", features = ["testing"] }
+tokio = { version = "1.42.0", features = ["full", "test-util"] }
 
 [[bench]]
 name = "request_processing"
--- a/src/core/mod.rs
+++ b/src/core/mod.rs
@@ -20,7 +20,7 @@ pub use circuit_breaker::{
 pub use error::{WorkerError, WorkerResult};
 pub use retry::{is_retryable_status, BackoffCalculator, RetryError, RetryExecutor};
 pub use worker::{
-    start_health_checker, BasicWorker, ConnectionMode, DPAwareWorker, HealthChecker, HealthConfig,
-    Worker, WorkerCollection, WorkerFactory, WorkerLoadGuard, WorkerType,
+    BasicWorker, ConnectionMode, DPAwareWorker, HealthChecker, HealthConfig, Worker,
+    WorkerCollection, WorkerFactory, WorkerLoadGuard, WorkerType,
 };
 pub use worker_registry::{WorkerId, WorkerRegistry, WorkerRegistryStats};
--- a/src/core/worker.rs
+++ b/src/core/worker.rs
@@ -57,9 +57,10 @@ pub trait Worker: Send + Sync + fmt::Debug {
     /// Decrement the load counter
     fn decrement_load(&self);
 
-    /// Reset the load counter to 0. Recovery-only: called exclusively when a
-    /// worker transitions back to healthy, never periodically. Health checks
-    /// must not discard active request accounting.
+    /// Reset the load counter to 0. Administrative operation: only safe when
+    /// the caller has independently established that no requests are in
+    /// flight for this worker. The runtime never calls this automatically;
+    /// health checks and routing must not discard active request accounting.
     fn reset_load(&self) {
         // Default implementation - does nothing
         // Workers that track load should override this
@@ -468,10 +469,9 @@ impl Worker for BasicWorker {
         RouterMetrics::set_worker_load(self.url(), self.load());
     }
 
-    /// Recovery-only: clears the load counter after the worker transitions
-    /// back to healthy. While unhealthy no new requests are routed to this
-    /// worker, so any remaining load is drift from the down period. Never
-    /// called periodically.
+    /// Administrative operation: resets the load counter to 0. Only safe when
+    /// the caller has independently established that no requests are in
+    /// flight for this worker; the runtime never calls this automatically.
     fn reset_load(&self) {
         self.load_counter.store(0, Ordering::Relaxed);
         RouterMetrics::set_worker_load(self.url(), 0);
--- a/src/core/worker.rs
+++ b/src/core/worker.rs
@@ -1,7 +1,6 @@
 use super::{CircuitBreaker, CircuitBreakerConfig, WorkerError, WorkerResult};
 use crate::metrics::RouterMetrics;
 use async_trait::async_trait;
-use futures;
 use serde_json;
 use std::fmt;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
@@ -58,7 +57,9 @@ pub trait Worker: Send + Sync + fmt::Debug {
     /// Decrement the load counter
     fn decrement_load(&self);
 
-    /// Reset the load counter to 0 (for sync/recovery)
+    /// Reset the load counter to 0. Recovery-only: called exclusively when a
+    /// worker transitions back to healthy, never periodically. Health checks
+    /// must not discard active request accounting.
     fn reset_load(&self) {
         // Default implementation - does nothing
         // Workers that track load should override this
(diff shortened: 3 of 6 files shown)


hsliuustc0106:
@ywang96 @zhuohan123 PTAL

bvolpato (changes requested):
Thanks for covering both cancellation and periodic-reset paths. One blocking streaming retry case remains in `send_typed_request`: retryable response headers call `guard.release()` before the forwarding task finishes consuming the backend body. The retry can start while the previous attempt is still draining or stalled, but that active attempt is already counted as zero. Cache-aware routing can then select the same busy worker and recreate the undercounting feedback loop.

Could we keep the unreleased guard in the forwarding task for retryable responses too, releasing it only when that stream ends or is cancelled? A focused test with retryable headers plus a held response body would protect this branch. Everything else in the accounting and P/D cancellation changes looks good.

bvolpato:
Could we combine receiver-closure handling from #215 and decode-body lifetime from #176 so cancellation also covers stalled bodies?

At `9e6fce2`, the regular streaming task still waits only on `stream.next()` (`router.rs:902`). Dropping the client response does not wake it until another upstream chunk arrives, so a stalled stream retains the guard and worker connection. In the PD path, `decode_guard.release()` still runs at response headers (`vllm_pd_router.rs:1471`), before either buffering or forwarding the body.

A headers-then-stall mock would cover both cases: close the downstream response and expect prompt release; while it stays connected, keep decode load counted until EOF/drop. #200 overlaps this work and has the same remaining boundaries, so consolidating these pieces would avoid another competing fix.

PR #217, opened 2026-08-16
[Feature]: cache-aware chat routing with session-id/full-history fallback

wuhang2014:
## Purpose

Cache-aware routing for `/v1/chat/completions` the way a real agent session works — sticky by `session_params.session_id` when that is enough, fall back to prefix-tree match on the full chat history when it is not, and still load-balance when workers are far enough apart.

For each chat request the `cache_aware` policy now receives two routing keys:

- **Primary**: the raw `session_params.session_id`, matched by **exact string equality** in a dedicated per-model session map (no prefix semantics, so `session-1` can never collide with `session-12`). Session entries are aged by last access (TTL sweep) and capped per model.
- **Fallback**: the serialized chat history (name-sorted tool schemas first, then system/developer/user/assistant/tool messages). Probed as a prefix-tree key at the configured `--cache-threshold` when the session map misses.

A session miss therefore still lands on the worker whose KV cache holds the conversation prefix; a miss on both keys falls back to min-load. If workers are imbalanced past the abs/rel gates, affinity is broken for min-load as before. This behavior is hardcoded (no new config knobs).

Additional behavior:

- Per-request decision labels (`session_id_match`, `full_history_match`, `load_balance`, `empty_history_min_load`, …) are emitted to `vllm_router_cache_aware_decisions_total{decision=…}` (exactly one final decision per request) and echoed in `x-vllm-router-decision` response headers alongside `x-vllm-router-worker` / `x-vllm-router-base-worker` / `x-vllm-router-dp-rank` (streaming and non-streaming).
- Stale-entry cleanup: dead tree tenants and dead session-map entries are dropped lazily on lookup and eagerly on worker removal.
- New `developer` chat role, with strict string-or-text-parts content: array-formatted content (text-only parts) is forwarded intact, non-text parts are a hard 4xx, mirroring the existing strict system-message handling.
- The min-load (imbalanced) path also teaches the session map and the prefix tree the full chat history, so a later balanced request can discover prefixes learned during load spikes.
- Session hits teach the tree the grown conversation history, so clients that later drop the session id still match the current prefix.

Compatibility notes:

- Hash-based policies (`consistent_hash`, `rendezvous_hash`) keep the pre-change routing text (raw session id); only `cache_aware` sees the session map and fallback.
- `SystemContentPart::Text` preserves unknown extension fields (e.g. `prompt_cache_breakpoint`) via `#[serde(flatten)]` across the parse/reserialize forwarding round-trip.

## Test Plan

```bash
cargo fmt --check
cargo clippy --all-targets --all-features -- -D warnings
cargo test --lib --bins
cargo test --test '*'
python -m black --check py_src/ py_test/
python -m ruff check py_src/ py_test/
pytest py_test/unit py_test/integration --ignore=py_test/e2e
```

New tests:

- `src/policies/cache_aware.rs`: exact-match session affinity + full-history fallback selection, empty-history session semantics (`empty_history_min_load`), imbalanced path teaches session map + full history, stale session-entry cleanup on worker removal, session-entry TTL/capacity eviction, header-key fallback.
- `src/protocols/spec.rs`: developer array-content round-trip + non-text-part rejection, system text-part extension-field preservation, raw session-id routing key.
- `src/routers/http/router.rs`: `chat_routing_keys` gives `cache_aware` the session key + fallback while hash policies keep the raw session id.
- `tests/test_prompt_input.rs`: full-history routing-key fixtures (tool sorting, volatile turns, array developer content).
- `py_test/integration/load_balancing/test_cache_aware.py`: stable agent prefix keeps chat affinity on one worker.

## Test Result

```text
cargo test --lib --bins     500 passed; 0 failed
cargo test --test '*'       all suites passed (test_openai_router_health 503 is a pre-existing macOS-only failure, also present on pristine main)
cargo clippy --all-targets --all-features -- -D warnings   clean
python -m black --check / ruff check   clean
pytest py_test/unit/test_arg_parser.py   20 passed, 1 skipped
```

Mutation probes: replacing the exact-match session map lookup with the old prefix-tree probe makes `test_cache_aware_empty_history_uses_exact_session_match` route `session-12` onto the `session-1` worker (shared string prefix); removing the history insert from the min-load path makes `test_cache_aware_imbalanced_path_teaches_full_history` fail.

Not yet run in this environment: `py_test/integration` (needs the rebuilt PyO3 extension) and a GPU smoke (`pip install -e .` after building).

### Performance evaluation

Eval: Qwen3.5-4B, Codex SWE-bench Pro cold-transcript replay (**25 sessions × 4 turns**, 100 requests), `max_tokens=256`, `session_serial` fire mode. **DP baseline** (data-parallel vLLM without the router) vs the router with the cache-aware chat routing and the `lb_mid` preset (`cache=0.3, abs=2, rel=1.5`). Cold start per case; `--enable-prefix-caching` on every backend. `Prompt%` is the prefix-cache hit rate; queue/prefill/decode are Prometheus means; TTFT/E2E are client means.

**NPU (Ascend, prefix block 1024), `max_num_batched_tokens=2048`**

| C | Case | Prompt% | Queue s | Prefill s | Decode s | TTFT s | E2E s |
|--:|------|--------:|--------:|----------:|---------:|-------:|------:|
| 4 | DP baseline | 20.5 | 0.27 | 2.59 | 7.80 | 2.95 | 10.74 |
| 4 | cache-aware | **59.5** | **0.17** | **1.35** | **5.25** | **1.64** | **6.87** |
| 8 | DP baseline | 20.7 | 1.40 | 2.84 | 11.84 | 4.41 | 16.23 |
| 8 | cache-aware | **50.1** | **0.99** | **1.82** | **7.87** | **2.97** | **10.82** |

C4: TTFT **−1.31 s**, E2E **−3.87 s**. C8: TTFT **−1.44 s**, E2E **−5.40 s**.
DP baseline Prompt% collapses to ~21% at `bt=2048`; session-id + full-history fallback restores ~50–60% and improves every measured column.

**GPU (H800), C4 / `max_tokens=256` / `bt=2048`**

| C | Case | Prompt% | Queue s | Prefill s | Decode s | TTFT s | E2E s |
|--:|------|--------:|--------:|----------:|---------:|-------:|------:|
| 4 | DP baseline | 38.0 | 0.88 | 7.66 | 18.24 | 8.71 | 26.93 |
| 4 | cache-aware | **50.1** | 0.85 | **6.40** | **14.25** | **7.46** | **21.70** |

Prompt/prefill/decode/TTFT/E2E all improve; queue is essentially unchanged (0.88 s vs 0.85 s mean). Client fire mode matters: these numbers assume `session_serial` — turn *n+1* waits for turn *n*, the serving shape of a real agent session.

---
<details>

</details>

### DIFF
--- a/benches/request_processing.rs
+++ b/benches/request_processing.rs
@@ -4,7 +4,7 @@ use std::time::Instant;
 
 use vllm_router_rs::protocols::spec::{
     ChatCompletionRequest, ChatMessage, CompletionRequest, GenerateParameters, GenerateRequest,
-    PromptInput, SamplingParams, UserMessageContent,
+    PromptInput, SamplingParams, SystemMessageContent, UserMessageContent,
 };
 
 /// Create a default GenerateRequest for benchmarks with minimal fields set
@@ -151,7 +151,7 @@ fn create_sample_chat_completion_request() -> ChatCompletionRequest {
         messages: vec![
             ChatMessage::System {
                 role: "system".to_string(),
-                content: "You are a helpful assistant".to_string(),
+                content: SystemMessageContent::Text("You are a helpful assistant".to_string()),
                 name: None,
             },
             ChatMessage::User {
@@ -192,7 +192,9 @@ fn create_sample_completion_request() -> CompletionRequest {
 fn create_large_chat_completion_request() -> ChatCompletionRequest {
     let mut messages = vec![ChatMessage::System {
         role: "system".to_string(),
-        content: "You are a helpful assistant with extensive knowledge.".to_string(),
+        content: SystemMessageContent::Text(
+            "You are a helpful assistant with extensive knowledge.".to_string(),
+        ),
         name: None,
     }];
 
--- a/docs/load_balancing/README.md
+++ b/docs/load_balancing/README.md
@@ -201,21 +201,23 @@ The `cache_aware` policy optimizes for prefix caching by maintaining an approxim
 
 ```bash
 vllm-router --policy cache_aware \
-  --cache-threshold 0.5 \
-  --balance-abs-threshold 32 \
-  --balance-rel-threshold 1.1 \
-  --worker-urls http://worker1:8000,http://worker2:8000
+  --cache-threshold 0.3 \
+  --balance-abs-threshold 64 \
+  --balance-rel-threshold 1.5 \
+  --worker-urls http://worker1:8000 http://worker2:8000
 ```
+
+Chat routing key is always session id first, then full chat history fallback.
 
 ### Parameters
 
 | Parameter | Default | Description |
 |-----------|---------|-------------|
-| `cache_threshold` | 0.5 | Minimum prefix match ratio to use cache-based routing |
-| `balance_abs_threshold` | 32 | Absolute load difference threshold for load balancing |
-| `balance_rel_threshold` | 1.1 | Relative load ratio threshold for load balancing |
-| `eviction_interval_secs` | 30 | Interval for cache eviction |
-| `max_tree_size` | 10000 | Maximum nodes per radix tree |
+| `cache_threshold` | 0.3 | Minimum prefix match ratio to use cache-based routing |
+| `balance_abs_threshold` | 64 | Absolute load difference that allows breaking affinity |
+| `balance_rel_threshold` | 1.5 | Relative load ratio that allows breaking affinity |
+| `eviction_interval_secs` | 120 | Interval for cache eviction |
+| `max_tree_size` | 2^26 | Maximum nodes per radix tree |
 
 ### Behavior
 
@@ -231,6 +233,7 @@ vllm-router --policy cache_aware \
 ### Best For
 
 - Workloads with repeated prompt prefixes (system prompts, few-shot examples)
+- Multi-turn chat / agent sessions when prefix caching is enabled on workers
 - When prefix caching is enabled on vLLM workers
 - Multi-tenant deployments with distinct prompt patterns
 
@@ -260,8 +263,8 @@ vllm-router --policy cache_aware \
                  │                       │       │                       │
                  ▼                       ▼       ▼                       ▼
         ┌────────────────┐     ┌────────────────┐ ┌────────────────┐ ┌────────────────┐
-        │ consistent_hash│     │  cache_aware   │ │  power_of_two  │ │  round_robin   │
-        │                │     │                │ │                │ │  or random     │
+        │ cache_aware    │     │ consistent_hash│ │  power_of_two  │ │  round_robin   │
+        │ (chat default) │     │                │ │                │ │  or random     │
         └────────────────┘     └────────────────┘ └────────────────┘ └────────────────┘
 ```
 
@@ -269,7 +272,8 @@ vllm-router --policy cache_aware \
 
 | Scenario | Recommended Policy |
 |----------|-------------------|
-| Chat applications with conversation history | `consistent_hash` |
+| Chat / multi-turn agent with prefix caching | `cache_aware` (default chat key: session then full history) |
+| Session stickiness without prefix matching | `consistent_hash` |
 | Batch inference with no state | `round_robin` |
 | Variable request complexity | `power_of_two` |
 | Repeated system prompts / few-shot | `cache_aware` |
--- a/examples/simulate_consistent_hash.rs
+++ b/examples/simulate_consistent_hash.rs
@@ -309,7 +309,7 @@ fn analyze_ring_distribution(
     for url in worker_urls {
         vnodes_per_url.insert(url.clone(), 0);
     }
-    for (_, url) in ring.iter() {
+    for url in ring.values() {
         *vnodes_per_url.get_mut(url).unwrap() += 1;
     }
 
--- a/py_src/vllm_router/router.py
+++ b/py_src/vllm_router/router.py
@@ -37,11 +37,11 @@ class Router:
         worker_startup_check_interval: Interval in seconds between checks for worker initialization. Default: 10
         cache_threshold: Cache threshold (0.0-1.0) for cache-aware routing. Routes to cached worker
             if the match rate exceeds threshold, otherwise routes to the worker with the smallest
-            tree. Default: 0.5
+            tree. Default: 0.3
         balance_abs_threshold: Load balancing is triggered when (max_load - min_load) > abs_threshold
-            AND max_load > min_load * rel_threshold. Otherwise, use cache aware. Default: 32
+            AND max_load > min_load * rel_threshold. Otherwise, use cache aware. Default: 64
         balance_rel_threshold: Load balancing is triggered when (max_load - min_load) > abs_threshold
-            AND max_load > min_load * rel_threshold. Otherwise, use cache aware. Default: 1.0001
+            AND max_load > min_load * rel_threshold. Otherwise, use cache aware. Default: 1.5
         eviction_interval_secs: Interval in seconds between cache eviction operations in cache-aware
             routing. Default: 60
         max_payload_size: Maximum payload size in bytes. Default: 256MB
--- a/py_test/integration/load_balancing/test_cache_aware.py
+++ b/py_test/integration/load_balancing/test_cache_aware.py
@@ -34,6 +34,62 @@ def test_cache_aware_affinity(mock_workers, router_manager):
 
 
 @pytest.mark.integration
+def test_cache_aware_chat_affinity_uses_stable_agent_prefix(
+    mock_workers, router_manager
+):
+    # Same system/tools but changing user suffix should stick to one worker.
+    _, urls, ids = mock_workers(n=2)
+    rh = router_manager.start_router(worker_urls=urls, policy="cache_aware")
+
+    tools = [
+        {
+            "type": "function",
+            "function": {
+                "name": "read_file",
+                "description": "Read a file",
+                "parameters": {
+                    "type": "object",
+                    "properties": {"path": {"type": "string"}},
+                },
+            },
+        }
+    ]
+
+    counts = collections.Counter()
+    with requests.Session() as s:
+        for i in range(12):
+            r = s.post(
+                f"{rh.url}/v1/chat/completions",
+                json={
+                    "model": "test-model",
+                    "messages": [
+                        {
+                            "role": "system",
+                            "content": "You are a coding agent.",
+                        },
+                        {
+                            "role": "developer",
+                            "content": "Inspect before editing.",
+                        },
+                        {
+                            "role": "user",
+                            "content": f"Fix changing task suffix {i}",
+                        },
+                    ],
+                    "tools": tools,
+                    "max_tokens": 1,
+                    "stream": False,
+                },
+            )
+            assert r.status_code == 200
+            wid = r.headers.get("X-Worker-Id") or r.json().get("worker_id")
+            counts[wid] += 1
+
+    top = max(counts.values())
+    assert top >= 10, counts
+
+
+@pytest.mark.integration
 def test_cache_aware_diverse_prompts_balances(mock_workers, router_manager):
     # Add latency so concurrent requests overlap and influence load-based selection
     _, urls, ids = mock_workers(n=3, args=["--latency-ms", "30"])
--- a/py_test/unit/test_arg_parser.py
+++ b/py_test/unit/test_arg_parser.py
@@ -440,7 +440,7 @@ class TestParseRouterArgs:
         assert router_args.worker_urls == ["http://worker1:8000", "http://worker2:8000"]
         assert router_args.policy == "round_robin"
 
-    def test_parse_pd_args(self):
+    def test_parse_pd_disaggregation_args(self):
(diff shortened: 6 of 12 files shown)


hsliuustc0106:
please provide an issue for requirement and design discussion

wuhang2014:
# RFC: Cache-aware chat routing with exact-match session affinity

- **Status:** Implemented — see [PR vllm-project/router#217](https://github.com/vllm-project/router/pull/217) (branch `feat/cache-aware-chat-routing`)
- **Authors:** WU Hang, herotai214
- **Date:** 2026-08
- **Related:** `cache_aware` policy, `/v1/chat/completions` routing

## Summary

Extend the `cache_aware` routing policy so `/v1/chat/completions` requests
route the way a real agent session works: sticky to a worker by
`session_params.session_id` (exact-match affinity), falling back to prefix-tree
matching on the full serialized chat history when the session is unknown, and
falling back to min-load when neither key has affinity. When workers are
imbalanced past the existing abs/rel gates, load balancing overrides affinity
as before.

Every routing decision is observable: a final decision label
(`session_id_match`, `full_history_match`, `load_balance`, …) is emitted to
Prometheus and echoed in the `x-vllm-router-decision` response header.

## Motivation

Agent workloads are multi-turn. Each turn sends the whole conversation to the
worker, and with vLLM prefix caching the KV cache pays for every token shared
with a previous turn on the same worker. On `main` (the baseline this RFC is
written against), `cache_aware` chat routing has three failure modes for this
shape:

1. **Session affinity is approximated by a prefix probe, not equality.**
   `main` passes `session_params.session_id` as the routing text and probes
   the radix tree at `cache_threshold` (default 0.5). "Same session" is
   therefore only approximate: two different sessions whose ids share a long
   string prefix (e.g. `session-1` vs `session-12`, 8/9 ≈ 89% match) count as
   the same affinity and pile onto one worker, while a session id that
   happens to prefix-match nothing gets no stickiness at all.
2. **Requests without a session id have no cache affinity.** On `main` the
   routing text is the session id and nothing else, so a chat request without
   `session_params.session_id` routes with an empty key — the tree probe
   always misses and the request goes to min-load, even when its system
   prompt and tool schemas are already cached on some worker.
3. **The conversation prefix is never modeled.** Even with a session id,
   cross-session prefix reuse (shared system prompt, tool schemas) is
   invisible to the policy, so the shared prefix is re-prefilled on every
   worker. There is also no observability of why a worker was chosen.

The evaluation below (Qwen3.5-4B, Codex SWE-bench Pro replay, 25 sessions × 4
turns) shows the DP baseline collapsing to ~21% prompt-cache hit rate while
session+history routing restores ~50–60% and improves TTFT/E2E on every
measured column.

## Goals

- `G1` Session affinity is exact string equality — never approximated via
  prefix thresholds.
- `G2` A request with an unknown session id still lands on the worker whose
  KV cache holds the conversation prefix.
- `G3` Load balancing still wins over affinity when workers are imbalanced.
- `G4` No new mandatory configuration knobs; behavior is on for chat requests
  under the `cache_aware` policy and invisible to every other policy.
- `G5` Per-request decision observability (metrics + response headers).

## Non-goals

- Token-level / occupancy-aware load accounting (still request-count).
- Modeling KV-cache eviction or per-block cache capacity on the router. The
  router's tree and session map are prefix-affinity estimates; worker
  `prompt_tokens_cached` remains the source of truth.
- Changing `consistent_hash` / `rendezvous_hash` semantics. They keep the raw
  session id with no fallback key.
- A configurable chat-key mode (an earlier iteration exposed
  `ChatRoutingKeyMode`; it was dropped in favor of one hardcoded behavior).

## Terminology

- **Session key**: a session identity derived with the shared
  `extract_hash_key` helpers — `session_params.session_id` for chat requests,
  or `header:<name>:<value>` / body `session:`/`user:` fields otherwise.
  Matched by exact equality.
- **History key**: canonical serialization of the chat body (see below).
  Probed through the prefix tree at `cache_threshold`.
- **Session map**: per-model `session_key → (worker_url, last_access_ms)`.
- **Decision label**: the final routing outcome, recorded once per request.

## Detailed design

### 1. Chat routing keys

`Router::chat_routing_keys(model, body)` resolves the policy for the model and
returns `(primary, fallback)`:

- `cache_aware`: `(session_id, Some(history_key))`
- every other policy: `(session_id, None)` — unchanged pre-RFC behavior, so
  hash-based affinities survive rolling upgrades.

`session_id` is `session_params.session_id`, trimmed; empty/absent yields an
empty primary key.

`history_key` is built by `ChatCompletionRequest::extract_full_history_routing_text()`:

1. Tool/function schemas first, sorted by name — schemas are stable across
   turns, so they stay in the matchable prefix while the message history
   grows. (`serde_json::Map` is BTreeMap-backed here, so parameter
   serialization is key-order canonical.)
2. Then, in order: `system:`, `developer:`, `user:` (text or text parts),
   `assistant:` content, `assistant_tool_call:name:args`,
   `assistant_function_call:name:args`, `tool:` content, `function:name:content`.

### 2. Session identity derivation

The policy derives the session key with the same helper hash policies use
(`extract_hash_key`, priority: HTTP headers → body fields → text), with one
deliberate deviation:

1. **Explicit routing text wins when non-empty.** For chat requests this is
   already the `session_params.session_id`; for other requests it is the
   prompt/history text.
2. Otherwise **HTTP headers** (`extract_hash_key_from_headers`:
   `x-session-id`, `x-user-id`, `x-tenant-id`, `x-correlation-id`,
   `x-request-id`, `x-trace-id`, formatted `header:<name>:<value>`).
3. Otherwise **body fields** (`extract_hash_key_from_body`:
   `session_params.session_id`, `user`, `session_id`, `user_id`).

A key is *session-derived* when it comes from source 2 or 3, or when a
fallback key marks the chat path (source 1 is then a session id by contract).

- Session-derived keys are matched by **exact string equality** in the
  session map — including header-only requests that carry no fallback key
  (e.g. an `x-session-id` header on a completions call).
- Plain text keys keep **prefix-tree semantics** even when unrelated headers
  are present (a prompt with an `x-request-id` header still gets prefix
  affinity, since per-request ids are unique and would otherwise defeat the
  prefix tree).

### 3. Session map (exact-match affinity)

`CacheAwarePolicy` keeps, next to the per-model prefix trees:

```text
session_keys: DashMap<model_id, DashMap<session_id, (worker_url, last_access_ms)>>
```

Session ids **never enter the prefix tree**. Lookup is `get` — exact string
equality. `session-1` and `session-12` are distinct keys with no possible
collision, regardless of how much prefix they share.

### 4. Selection algorithm

For a session-derived key (chat requests via the fallback contract, or
header/body-derived keys — see §2):

1. **Imbalance gate** (unchanged): if `max_load - min_load > abs_threshold`
   and `max_load > min_load * rel_threshold`, route min-load and skip the
   probes below (the min-load path still teaches cache state, see §5).
2. **Session lookup**:
   - Exact hit and the mapped worker is present and healthy →
     decision `session_id_match`.
   - Exact hit but the worker is gone/unhealthy → drop the stale entry and
     proceed to the history probe.
   - No entry → proceed to the history probe.
3. **History probe**: prefix-match `history_key` at `cache_threshold`
   (strict `>`):
   - hit → `full_history_match`; hit on a dead tenant →
     `stale_tenant_fallback` (first healthy worker);
   - miss → `full_history_low_match` (min-load).
4. **No usable history** (no fallback key provided, or empty extracted
   history such as an image-only chat): session miss selects min-load
   directly with decision `empty_history_min_load` — never a prefix probe
   of an empty key.
5. Requests without a session-derived key (plain text, also used by non-chat
   traffic) keep the pre-RFC single prefix probe at `cache_threshold`
   (`cache_affinity` / `low_match_min_load`).
6. No tree for the model: `no_tree_random`.

### 5. Teaching cache state

Every selection teaches the state that produced it:

- On a session hit: refresh the session entry timestamp and insert the
  (grown) history key into the tree, so clients that later drop the session
  id still match the current prefix.
- On a history hit / min-load / imbalanced selection: set
  `session_map[session_key] = selected worker` and, when a history key
  exists, insert it into the tree for the selected worker.
- Header-only session requests (no fallback key) teach only the session
  map; the min-load (imbalanced) path applies the same session+history
  inserts as the balanced path.

### 6. Session map eviction

Entries carry `last_access_ms` and are maintained by the existing eviction
cycle (and `evict_cache`):

- **TTL**: entries untouched for `SESSION_ENTRY_TTL_SECS` (1 h) are removed.
- **Capacity**: if a model's map exceeds `max_tree_size` entries, the least
  recently accessed entries are evicted first.
- **Worker removal**: `remove_worker` / `remove_worker_by_url` eagerly drop
  all entries pointing at the removed URL; lookups lazily drop entries whose
  worker is no longer healthy.

### 7. Protocol surface

- `ChatCompletionRequest::extract_session_id_for_routing()` returns the raw
  session id; `extract_text_for_routing()` returns the full-history key with
  a raw session-id fallback.
- New `ChatMessage::Developer` variant with strict string-or-text-parts
  content (`SystemMessageContent` shape): array content with text-only parts
  round-trips; non-text parts are a hard 4xx; missing/null → empty text.
- `SystemContentPart::Text` carries a `#[serde(flatten)]` map so extension
  fields (e.g. `prompt_cache_breakpoint`) survive parse/reserialize
  forwarding.

### 8. Observability

- `vllm_router_cache_aware_decisions_total{decision=…}` — exactly one
  increment per request, with the final decision label.
- Response headers: `x-vllm-router-decision` plus
  `x-vllm-router-worker` / `x-vllm-router-base-worker` /
  `x-vllm-router-dp-rank` (streaming and non-streaming).

## Alternatives considered

- **Keep the `main` single prefix probe on the raw session id**: rejected —
  prefix matching is an approximation of equality (prefix-related session ids
  collide above `cache_threshold`) and an absent session id has no affinity
  at all, which is exactly the agent-workload failure mode this RFC fixes.
- **Reuse `consistent_hash` for sessions**: sticky, but has no history
  fallback and would require policy switching mid-request; the session map is
  also warm-start friendly (cheap to rebuild, no ring).
- **Token-level / occupancy load**: out of scope; request-count balancing is
  unchanged.

## Performance evaluation

Qwen3.5-4B, Codex SWE-bench Pro cold-transcript replay (25 sessions × 4 turns,
100 requests), `max_tokens=256`, `session_serial` fire mode; DP baseline vs
the router with `lb_mid` preset (`cache=0.3, abs=2, rel=1.5`), cold start per
case, prefix caching enabled.

| Device | C | Case | Prompt% | Queue s | Prefill s | Decode s | TTFT s | E2E s |
|---|---|------|--------:|--------:|----------:|---------:|-------:|------:|
| NPU (bt=2048) | 4 | DP baseline | 20.5 | 0.27 | 2.59 | 7.80 | 2.95 | 10.74 |
| NPU (bt=2048) | 4 | cache-aware | **59.5** | **0.17** | **1.35** | **5.25** | **1.64** | **6.87** |
| NPU (bt=2048) | 8 | DP baseline | 20.7 | 1.40 | 2.84 | 11.84 | 4.41 | 16.23 |
| NPU (bt=2048) | 8 | cache-aware | **50.1** | **0.99** | **1.82** | **7.87** | **2.97** | **10.82** |
| GPU H800 | 4 | DP baseline | 38.0 | 0.88 | 7.66 | 18.24 | 8.71 | 26.93 |
| GPU H800 | 4 | cache-aware | **50.1** | 0.85 | **6.40** | **14.25** | **7.46** | **21.70** |

C4: TTFT −1.31 s, E2E −3.87 s (NPU). C8: TTFT −1.44 s, E2E −5.40 s (NPU).
GPU queue time is essentially unchanged (0.88 s vs 0.85 s).

## Compatibility and migration

- Hash-based policies see the exact same routing text as before (raw session
  id), so rolling upgrades do not remap existing sessions.
- `cache_aware` chat behavior changes by design (session stickiness now exact;
  history fallback is new), but only for requests routed by `cache_aware`.
- No new required configuration; `cache_threshold`, balance gates and
  `max_tree_size` are reused. Session-map TTL is a compile-time constant.

## Testing plan

- Unit: exact-match session affinity incl. prefix-related ids, session hit
  teaches grown history, empty-history semantics, imbalanced path teaching,
  stale entry cleanup on worker removal, session-entry TTL/capacity eviction,
  header-only session affinity without a fallback key, text keys keeping
  prefix semantics when headers are present, header-key fallback, routing-key
  construction (schemas-before-history, tool sorting, developer arrays),
  extension-field preservation round-trips.
- Router-level: `chat_routing_keys` policy split (cache_aware vs hash
  policies).
- Integration: `py_test/integration/load_balancing/test_cache_aware.py`
  (stable agent prefix affinity with mock workers).
- E2E (GPU): the performance evaluation above; smoke =
  `benchmarks/run_dp_cache_aware_demo.sh` with prompt/APC hit-rate checks.

## Open questions

- Should the 1 h session TTL become a CLI flag once multi-day sessions are a
  real workload?

## Checklist

- [x] Behavior implemented and reviewed (exact-match redesign incorporated)
- [x] Tests (unit + integration) added for all new paths
- [x] Metrics and tracing headers documented
- [x] Performance evaluation captured in the PR description
- [ ] GPU smoke re-run on the final exact-match design

PR #221, opened 2026-08-20
[Bugfix]: strip @rank suffix in DP intra-node GET proxy URLs

### LINKED ISSUE #220: [Bug] DP intra-node GET proxies fail when worker URLs contain @rank
**Title:** `[Bug] DP intra-node GET proxies fail when worker URLs contain @rank (userinfo misinterpretation)`

**Description**

When running with intra-node data parallel size > 1, worker registry URLs carry an `@rank` suffix (e.g. `http://127.0.0.1:18100@0`). The GET proxy path in `proxy_get_request` (`src/routers/http/router.rs`) forwards these URLs without stripping `@rank`:

```rust
let url = format!("{}/{}", worker_url, endpoint);   // http://127.0.0.1:18100@0/v1/models
```

Per RFC 3986, `@` in the authority part is parsed as userinfo, so reqwest sends this request to host `0` instead of `127.0.0.1:18100`. The request fails and the router returns 500.

**Affected endpoints (all via `proxy_get_request`)**
- `GET /v1/models` (breaks Swarm / AgentBench preflight)
- `GET /health_generate`
- `GET /get_server_info`
- `GET /get_model_info`

Unaffected: `POST /v1/chat/completions` (already strips `@rank` via `dp_utils::parse_worker_url` and sends `X-data-parallel-rank`) and `GET /health` (`send_health_check` already strips via `extract_dp_rank`). DP=1 is unaffected since there is no `@rank` suffix.

**Reproduction** (uses the repo's own mock worker; verified with the release binary)

```bash
# 1. Start the repo's mock worker (py_test/fixtures/mock_worker.py)
python py_test/fixtures/mock_worker.py --host 127.0.0.1 --port 23340 > worker.log 2>&1 &

# 2. Start the router with DP=2 (creates workers http://127.0.0.1:23340@0 and @1)
RUST_LOG=vllm_router_rs=debug target/release/vllm-router \
  --host 127.0.0.1 --port 8001 \
  --worker-urls http://127.0.0.1:23340 \
  --policy round_robin \
  --intra-node-data-parallel-size 2 \
  --health-check-interval-secs 1 \
  --prometheus-port 7293 \
  > router.log 2>&1 &

# 3. Wait for "Router ready", then hit the preflight endpoint:
curl -sS -i http://127.0.0.1:8001/v1/models
```

Observed output:

```
# router.log
Router ready | workers: ["http://127.0.0.1:23340@0", "http://127.0.0.1:23340@1"]

# GET /v1/models through router (DP=2)
HTTP/1.1 500
Request failed: error sending request for url (http://0.0.0.0/v1/models)

# contrast 1 — direct GET to worker:                       200
curl http://127.0.0.1:23340/v1/models
{"data":[{"id":"mock","object":"model"}]}

# contrast 2 — POST through router (DP=2):                 200
curl -X POST http://127.0.0.1:8001/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{"model":"mock","messages":[{"role":"user","content":"hi"}]}'
{"id":"cmpl-...","choices":[{"text":"ok",...}],"worker_id":"worker-23340",...}

# contrast 3 — GET through router with DP=1:               200
```

The 500 body shows the outbound URL became `http://0.0.0.0/v1/models` — the router sent the request to host `0` (port 80) instead of `127.0.0.1:23340`; the worker never receives it. POST and DP=1 paths succeed, isolating the bug to the `@rank` suffix in `proxy_get_request`.

**Root cause**

`select_first_worker()` returns `workers[0].url()`, which for `DPAwareWorker` includes the `@rank` suffix (kept for registry identification; `base_url()` is the clean variant). `proxy_get_request` builds the outbound URL from the raw value.

**Suggested fix**

Parse the worker URL before building the request, and set the DP rank header the same way the chat path does:

```rust
let (base_url, dp_rank) = dp_utils::parse_worker_url(&worker_url);
let url = format!("{}/{}", base_url, endpoint);
let mut request_builder = dp_utils::add_dp_rank_header(self.client.get(&url), dp_rank);
```

Add a regression test for DP=2: `GET /v1/models` must return 200 through the router.

wuhang2014:
Fix #220: intra-node DP GET proxies fail when worker URLs contain `@rank`.

When `--intra-node-data-parallel-size > 1`, worker registry URLs carry an `@rank` suffix (e.g. `http://127.0.0.1:18100@0`). `proxy_get_request` forwarded these URLs to reqwest without stripping `@rank`, so reqwest parsed `@` as RFC 3986 userinfo and sent the request to the rank digit as host instead of the worker (e.g. `GET /v1/models` became a request to `http://0.0.0.0/v1/models` and returned 500). Chat POST already stripped `@rank`; GET did not.

## Purpose

Fix #220 (worker URLs with `@rank` are misinterpreted as userinfo in intra-node DP GET proxies).

- `proxy_get_request` now parses the worker URL with `dp_utils::parse_worker_url` and sets `X-data-parallel-rank` via `dp_utils::add_dp_rank_header`, matching the chat POST path. DP=1 is unchanged (no `@rank`, no extra header).
- Affected endpoints fixed: `GET /v1/models`, `GET /health_generate`, `GET /get_server_info`, `GET /get_model_info`.
- `GET /health` was already fine (`send_health_check` strips `@rank`).
- Added `test_regular_router_dp2_get_v1_models` in `tests/test_dp_routing.rs`: asserts `GET /v1/models` through a DP=2 regular router returns 200 with the worker's model list. Verified the test fails on the previous implementation (500 from the userinfo-corrupted URL) and passes with the fix.

## Test Plan

- [x] `cargo test --test test_dp_routing test_regular_router_dp2_get_v1_models`
- [x] `cargo test --test test_dp_routing`
- [x] `cargo test --test test_transparent_proxy_routing --test api_endpoints_test`
- [x] `cargo test --lib --bins`
- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`

## Test Result

```
running 1 test
test dp_e2e_tests::test_regular_router_dp2_get_v1_models ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 16 filtered out; finished in 0.69s
```

```
test result: ok. 17 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.39s
```

```
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

```
test result: ok. 484 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 75.19s
```

Before the fix, the regression test fails (worker never receives the request):

```
thread 'dp_e2e_tests::test_regular_router_dp2_get_v1_models' panicked:
GET /v1/models must succeed when DP > 1 (got 500). @rank in the worker URL was misinterpreted as userinfo.
```

---
<details>

</details>


fake0fan:
Thanks for the fix! Parsing the worker URL before constructing the GET request looks correct and addresses #220. Preserving the selected worker’s DP rank in the header also makes sense, even though endpoints such as `/v1/models` do not consume it. The regression test checks both the status and the worker’s model list, which provides good coverage of the original failure.

I have two suggestions before merging:

1. **Verify the forwarded DP rank.** Could we extend the mock worker’s `/v1/models` handler to capture request headers and assert that `X-data-parallel-rank` matches the selected worker’s rank? Currently, removing `add_dp_rank_header()` would still leave this test passing.

2. **Handle a conflicting client-provided rank header.** The router adds its selected rank before copying incoming headers, and that loop does not exclude `X-data-parallel-rank`. This could introduce duplicate values when the client supplies its own rank. When the router selects a DP rank, could we filter out the incoming value and then inject the selected rank?

It would be useful to cover the second case with a conflicting client header and assert that the worker receives **exactly one** rank value matching the router’s selection. That assertion should inspect all values, since the existing capture helper stores headers in a `HashMap`, which cannot preserve duplicates.

wuhang2014:
Thanks @fake0fan! Both suggestions are addressed in 9480f7e:

1. **Verify the forwarded DP rank:** the mock worker's `/v1/models` handler now captures request headers, and `test_regular_router_dp2_get_v1_models` asserts the request carried `X-data-parallel-rank` with a value matching one of the deployment's ranks (0 or 1 — both DP ranks share the mock worker, and `select_first_worker` order over the registry is not deterministic, so the assertion checks membership rather than a fixed rank). Removing `add_dp_rank_header()` now fails the test.

2. **Conflicting client-provided rank header:** `proxy_get_request` now skips the incoming `x-data-parallel-rank` when the router selected a DP rank, so the worker receives exactly one value (the router's). DP=1 passthrough is unchanged since the router injects nothing in that case. To make the assertion duplicate-proof, `CapturedRequest` now stores all header values (`HashMap<String, Vec<String>>`) instead of collapsing them. New test `test_regular_router_dp2_get_v1_models_overrides_client_dp_rank` sends `X-data-parallel-rank: 99` and asserts the worker received exactly one value equal to the router's selection; on the pre-fix code it fails with `worker received ["1", "99"]`.

All gates pass: `cargo test --test test_dp_routing` (18), otel suites, `--lib --bins` (484), `cargo fmt --check`, clippy `-D warnings`.

fake0fan (approved):
Thanks for addressing both points! The header filtering and duplicate-preserving capture cover the concerns I raised. The updated tests look sufficient for this fix. LGTM.

PR #224, opened 2026-08-27
fix(pd): close downstream stream at SSE done event

cquil11:
## Summary

- treat the OpenAI-compatible `data: [DONE]` SSE event as the protocol-level end of a P/D decode stream
- detect the terminal line across arbitrary HTTP chunk boundaries
- apply the wrapper to both discovered and direct vLLM P/D routing paths
- preserve normal upstream EOF and avoid matching marker text inside generated JSON payloads

## Motivation

A decode backend can finish inference and emit `[DONE]` while its HTTP body remains pending. Directly forwarding `reqwest::Response::bytes_stream()` then leaves the downstream client waiting indefinitely even though the backend has drained. Closing the downstream body at the OpenAI SSE terminal event makes the router follow the application protocol instead of depending on transport EOF.

## Validation

- regression test with `[DONE]` followed by an upstream stream that never reaches EOF
- split-marker chunk-boundary test
- payload false-positive test
- ordinary EOF test
- `cargo test --lib`: 488 passed

### DIFF
--- a/src/routers/http/vllm_pd_router.rs
+++ b/src/routers/http/vllm_pd_router.rs
@@ -18,10 +18,14 @@ use axum::{
     http::{HeaderMap, Method, StatusCode},
     response::{IntoResponse, Response},
 };
+use bytes::Bytes;
+use futures_util::Stream;
 use serde_json::{json, Value};
 use std::collections::HashMap;
+use std::pin::Pin;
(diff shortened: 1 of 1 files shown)


cquil11:
Closing because the project scope is now limited to porting active, nondeprecated AMD multi-node configurations to srt-slurm. This PR is outside that scope.

PR #225, opened 2026-08-28
[Bugfix]: Authenticate worker health checks

### LINKED ISSUE #159: [Bug]: `--api-key` is not applied to worker health checks
## Summary

`vllm-router` provides an `--api-key` option for worker authorization, but it does not appear to be applied to startup worker health checks.

When worker URLs require `Authorization: Bearer <token>`, the router stays at:

```text
Waiting for 3 unique hosts (representing 3 workers) to become healthy
````

and never reaches `Router ready`.

## Environment

* Router: `vllm-router`
* Deployment: Docker Compose on a VM
* Worker URLs: external HTTPS URLs for vLLM OpenAI-compatible endpoints
* Workers require Bearer token authentication

## Command

```bash
vllm-router \
  --host 0.0.0.0 \
  --port 8000 \
  --worker-urls \
    https://<worker-1> \
    https://<worker-2> \
    https://<worker-3> \
  --policy round_robin \
  --max-concurrent-requests 10000 \
  --api-key <BACKEND_API_KEY>
```

## Actual behavior

The router does not become ready:

```text
Waiting for 3 unique hosts (representing 3 workers) to become healthy
```

Direct requests to the same worker URLs work when the same API key is provided:

```bash
curl https://<worker-1>/health \
  -H "Authorization: Bearer <BACKEND_API_KEY>"
```

## Expected behavior

If `--api-key` is intended for worker authorization, it should also be used for worker health checks, for example:

```http
Authorization: Bearer <BACKEND_API_KEY>
```

## Impact

This prevents using `vllm-router` with external worker URLs protected by API key authentication, because startup health checks fail before the router can become ready.

## Question

Is `--api-key` expected to apply to worker health checks? If not, is there another supported option to set headers for worker health checks?

StevenWang-CY:
## Purpose

Fixes #159.

`--api-key` is defined as the credential used to authorize requests to workers. The regular router did not attach it to worker health requests. An authenticated worker therefore returned 401 during the startup gate, causing the router to time out before becoming ready. Health checks issued after registration had the same omission and could later mark the worker unhealthy.

This change applies the configured bearer token to every regular-router health path:

- the initial worker readiness gate;
- direct health forwarding and dynamic worker registration;
- recurring health checks performed through `BasicWorker`, including DP-aware workers.

Workers remain unauthenticated by default, and the existing public `wait_for_healthy_workers` signature retains its behavior. The credential is held in a private field and omitted from the worker's custom `Debug` output.

## Test Plan

The regression tests run an in-process Axum worker whose `/health` endpoint accepts only the exact configured bearer token. They cover initial startup, invalid-key rejection, direct health forwarding, dynamic registration, credential retention for recurring checks, and secret redaction. Existing tests continue to cover workers without authentication and DP-aware URL handling.

## Test Result

On the unmodified `main` commit, the authenticated-startup regression timed out waiting for the local worker. With this change, all local validation passed on macOS arm64 with Rust 1.95.0, matching the compiler version in Buildkite:

- `cargo fmt --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test --lib --bins`: 489 passed
- `cargo test --test '*'`: 180 passed
- `pre-commit run --files src/core/worker.rs src/routers/http/router.rs`

### DIFF
--- a/src/core/worker.rs
+++ b/src/core/worker.rs
@@ -284,6 +284,7 @@ pub struct WorkerMetadata {
 #[derive(Clone)]
 pub struct BasicWorker {
     metadata: WorkerMetadata,
+    api_key: Option<String>,
     load_counter: Arc<AtomicUsize>,
     processed_counter: Arc<AtomicUsize>,
     healthy: Arc<AtomicBool>,
@@ -322,6 +323,7 @@ impl BasicWorker {
 
         Self {
             metadata,
+            api_key: None,
             load_counter: Arc::new(AtomicUsize::new(0)),
             processed_counter: Arc::new(AtomicUsize::new(0)),
             healthy: Arc::new(AtomicBool::new(true)),
@@ -338,6 +340,12 @@ impl BasicWorker {
 
     pub fn with_health_config(mut self, config: HealthConfig) -> Self {
         self.metadata.health_config = config;
+        self
+    }
+
+    /// Configure the API key used to authenticate worker health checks.
+    pub fn with_api_key(mut self, api_key: Option<String>) -> Self {
+        self.api_key = api_key;
         self
     }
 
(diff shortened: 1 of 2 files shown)


StevenWang-CY:
@Prowindy, could you review and approve this fix for #159? It propagates the configured worker bearer token through startup readiness, direct and registered health forwarding, and recurring BasicWorker checks, while preserving unauthenticated defaults and redacting the credential from Debug output. DCO and Buildkite are green; local validation passed formatting, Clippy with warnings denied, 489 library and binary tests, and 180 integration tests. I will address any findings promptly.

PR #228, opened 2026-08-31
simplify CLI arg parsing

nt591:
## Purpose
Code legibility, general Rust-iness, and hopefully being useful to others.

Another driveby as I was working through the codebase. Today, CLI parsing walks the list of Clap args twice. Once for prefill, then another time for non-prefill args. We can do this once using standard iterator traversal which also simpified some of the index checks needed earlier.

This also fixes (I hope) a latent bug where `--prefill --port 3000` would silently drop instead of making noise about a missing URL.

## Test Plan
cargo check, cargo fmt, clippy, unit tests (new and old)

## Test Result
clean.
---
<details>

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,39 +9,46 @@ use vllm_router_rs::metrics::PrometheusConfig;
 use vllm_router_rs::server::{self, ServerConfig};
 use vllm_router_rs::service_discovery::ServiceDiscoveryConfig;
 
-// Helper function to parse prefill arguments from command line
-// Returns prefill_entries with (URL, optional_bootstrap_port)
-fn parse_prefill_args() -> Vec<(String, Option<u16>)> {
-    let args: Vec<String> = std::env::args().collect();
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -9,39 +9,45 @@ use vllm_router_rs::metrics::PrometheusConfig;
...
+
+        // optional port afterwards
+        let port = match args.peek() {
+            Some(arg) if arg.eq_ignore_ascii_case("none") => {
+                args.next(); // consume it and ignore
+                None
+            }
+            Some(arg) if !arg.starts_with("-") => {
+                // assume to be a port
+                let port = arg.parse::<u16>().ok();
```
nt591: this is intended to match with existing behavior, but in my opinion this branch should be something like `let port = arg.parse::<u16>().ok_or(format!("Invalid input for port: {arg:?}"))`

wuhang2014 (approved):
LGTM

PR #232, opened 2026-09-02
fix: skip unhealthy workers in proxy GET endpoints

### LINKED ISSUE #135: /v1/models returns 500 when any worker is unhealthy
When running the router with multiple `--worker-urls` and one of the workers is down, the `/v1/models` endpoint returns a 500 with a raw connection error. Chat completions work fine since they only get routed to healthy workers, but `/v1/models` seems to query all configured workers including unhealthy ones.

This is easy to reproduce: just point the router at two backends where one is not reachable. Chat completions will happily go to the healthy one, but `/v1/models` will fail every time.

In production this is a problem because workers restart, scale up/down, etc. and clients calling `/v1/models` will get 500s until every single backend is up. Ideally it should just return models from the healthy workers.

Let me know if you're interested in a PR.

Bevisy:
## Purpose

Fixes #135 — `/v1/models` (and other proxy GET endpoints) returns 500 when any worker is unhealthy.

`select_first_worker` returned the first registered worker without checking health state. `worker_registry.get_all()` returns all workers including unhealthy ones, so when the first worker was temporarily down, `proxy_get_request` proxied to a dead worker and returned 500. This affected `/v1/models`, `/get_server_info`, and `/get_model_info` — all of which route through `proxy_get_request` → `select_first_worker`.

Root-cause fix in the shared helper: filter for a healthy worker (`is_healthy()`) before returning; fall back to the existing error path when none are healthy.

## Test Plan

New unit tests in `routers::http::router::tests`:
- `test_select_first_worker_skips_unhealthy` — marks worker1 unhealthy, asserts worker2 is selected
- `test_select_first_worker_all_unhealthy` — marks all unhealthy, asserts error

## Test Result

```
running 3 tests
test routers::http::router::tests::test_select_first_worker_regular ... ok
test routers::http::router::tests::test_select_first_worker_skips_unhealthy ... ok
test routers::http::router::tests::test_select_first_worker_all_unhealthy ... ok

test result: ok. 3 passed; 0 failed
```

Built with `cargo test --lib routers::http::router::tests::test_select_first_worker` (rust:latest container, aarch64).

### DIFF
--- a/src/routers/http/router.rs
+++ b/src/routers/http/router.rs
@@ -338,11 +338,12 @@ impl Router {
     }
 
     fn select_first_worker(&self) -> Result<String, String> {
+        // Prefer a healthy worker so proxy GETs (e.g. /v1/models) don't 500
+        // when the first registered worker is temporarily down.
         let workers = self.worker_registry.get_all();
-        if workers.is_empty() {
-            Err("No workers are available".to_string())
-        } else {
-            Ok(workers[0].url().to_string())
+        match workers.into_iter().find(|w| w.is_healthy()) {
+            Some(worker) => Ok(worker.url().to_string()),
+            None => Err("No healthy workers are available".to_string()),
         }
     }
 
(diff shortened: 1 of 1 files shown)


Bevisy:
Handled in 4d03774.

> This can still select a worker whose circuit breaker is open: `is_healthy()` remains true in that state, while the repository defines availability as health plus `can_execute()` (`src/core/worker.rs:79-82`). Since this helper feeds the "first available worker" proxy path, use `is_available()` (and update the error text) so GET requests do not bypass a known-open circuit. A regression test with a healthy worker whose circuit is forced open would cover this case.

`select_first_worker` now uses `is_available()`, and the error text is updated to match ("No available workers are available"). Added `test_select_first_worker_skips_circuit_open`: worker1 stays healthy but with its circuit forced open, and is skipped in favor of worker2.

> "First registered worker" is inaccurate because `WorkerRegistry::get_all()` iterates a `DashMap`, whose order is unspecified (the existing test at `src/routers/http/router.rs:1828-1829` explicitly accounts for this). Describe this as an arbitrary registered worker so the comment matches the actual selection behavior.

Comment updated to describe an arbitrary registered worker (whose circuit may also be open).

PR #234, opened 2026-09-02
fix: accept all OpenAI reasoning_effort values

Bevisy:
## Purpose

Fixes #233 — the router rejects `reasoning_effort` values beyond `low`/`medium`/`high`.

`ReasoningEffort` (src/protocols/spec.rs) only defined `Low`/`Medium`/`High`, but OpenAI reasoning models accept `none`, `minimal`, `low`, `medium`, `high`, `xhigh`, and `max`. Because the router deserializes `reasoning_effort` into this enum with serde default, any request carrying the missing values failed deserialization and returned a 400 before reaching the backend — even when the backend worker supported the value.

The enum is shared by two paths, both affected:
- Chat Completions top-level `reasoning_effort` (spec.rs:522)
- Responses API `ResponseReasoningParam.effort` (spec.rs:848)

Fix: widen the enum to all seven OpenAI values. serde `rename_all = "snake_case"` already maps the new variants (`Xhigh`→`xhigh`, etc.). The router validates shape, not per-model value compatibility — that is the backend job.

## Test Plan

Extended `test_reasoning_effort_field` to cover all 7 values, and added `test_reasoning_effort_serde_roundtrip_all_values` which asserts each value deserializes from its snake_case string and re-serializes to the same token — directly covering the regression (previously `max`/`xhigh`/`none`/`minimal` 400'd at deserialization).

## Test Result

```
running 18 tests
test protocols::validation::tests::chat_tests::test_reasoning_effort_field ... ok
test protocols::validation::tests::chat_tests::test_reasoning_effort_serde_roundtrip_all_values ... ok
...
test result: ok. 18 passed; 0 failed; 0 ignored
```

Built with `cargo test --lib protocols::validation` (rust:latest container, aarch64).

PR #239, opened 2026-09-04
fix: accept current vLLM reasoning_effort values

HuXinjing:
## Summary

- accept the reasoning effort values currently supported by vLLM: `none`, `minimal`, `low`, `medium`, `high`, `xhigh`, and `max`
- preserve the selected value when forwarding requests
- cover both Chat Completions and Responses request schemas with serialization round-trip tests

## Motivation

The router currently rejects `reasoning_effort: "none"` during request deserialization with HTTP 422, before the request can reach a vLLM backend. Current vLLM uses `none` to disable thinking for supported models and exposes the expanded set of effort values in its OpenAI-compatible protocol.

## Validation

- `cargo test --test test_openai_routing reasoning_effort_accepts_current_vllm_values`
- `cargo clippy --all-targets --all-features -- -D warnings`

Validated separately against a vLLM-backed deployment: a Chat Completions request containing `reasoning_effort: "none"` returned HTTP 200.

### DIFF
--- a/src/protocols/spec.rs
+++ b/src/protocols/spec.rs
@@ -517,7 +517,7 @@ pub struct ChatCompletionRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub echo: Option<bool>,
 
-    /// Reasoning effort level for reasoning models (low, medium, high)
+    /// Reasoning effort level supported by OpenAI-compatible vLLM servers.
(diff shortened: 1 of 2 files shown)


wuhang2014:
This is a duplicate of #234, so I'll close it. Please reopen it if there is anything else to fix.

PR #246, opened 2026-09-08
[CI]: Fix CI failures from EOL bullseye base images

wuhang2014:
## Description

Debian 11 (bullseye) LTS ended on 2026-08-31 and the `bullseye-security` repository is no longer refreshed. Its `InRelease` metadata has now expired, so `apt-get update` fails inside any bullseye-based image:

```
E: Release file for http://deb.debian.org/debian-security/dists/bullseye-security/InRelease
is expired (invalid since 11h 24min 17s). Updates for this repository will not be applied.
```

This broke the `:python: Build Python Wheel` Buildkite job (`apt-get update && apt-get install python3-pip` chain stops → `pip3: command not found`), e.g. [router build #773](https://buildkite.com/vllm/router/builds/773), with all downstream jobs failing as `waiting_failed`. The same expired repos also broke the Trivy GHA job at the `docker build` step on every branch, including `main`.

## Changes

**1. Move off EOL bullseye (`57185c9`, `0263c2c`)**
- `.buildkite/pipeline.yml`: `RUST_IMAGE` `rust:1.95.0-bullseye` → `rust:1.95.0-bookworm`
- `Dockerfile.router`: `rustlang/rust:nightly-bullseye` → `nightly-bookworm`; `python:3.12-slim-bullseye` → `python:3.12-slim-bookworm`
- `.buildkite/release-pipeline.yml`: smoke-test the release binary on `debian:bookworm` (binary is built on manylinux_2_28 / glibc 2.28, still within bookworm's glibc 2.36)

**2. Use a venv for pip on bookworm (`dd9603f`)**
Debian 12 marks the system Python as externally managed (PEP 668), so bare `pip3 install` fails with `externally-managed-environment` ([build #778](https://buildkite.com/vllm/router/builds/778)). The wheel-build and python-test jobs now install `python3-venv` (+ explicit `python3-dev` for the PyO3 build) and run `pip`/`python`/`pytest` inside a venv. Also uses `apt-get upgrade` instead of `apt upgrade` in `Dockerfile.router` per review (apt's CLI is not stable for scripting).

**3. Drop pip from the runtime image (`2416eb2`)**
The Trivy scan of the built image flagged pip's own vendored `msgpack` 1.1.2 (GHSA-6v7p-g79w-8964, HIGH) and `setuptools` 70.3.0 (CVE-2025-47273 HIGH, CVE-2026-59890 MEDIUM). These live inside pip's `_vendor` tree, so no surrounding package change can clear them, and pip will not bump the vendored setuptools ([pypa/pip#14031](https://github.com/pypa/pip/issues/14031)). The runtime image only needs `requests` for the optional Python launcher; the CMD runs the Rust binary directly. The final stage now installs `requests`, then uninstalls pip entirely and removes `ensurepip`'s bundled wheel (which carries the same vendored copies). This also drops the `setuptools-rust`/`wheel`/`build` build tools from the image.

## Verification

- Buildkite pre-merge pipeline passes on this branch (wheel build, python tests included)
- Trivy GHA job passes on this branch (`2416eb2`) — previously failing on all branches, including `main`
- Replacement image tags all exist on Docker Hub: `rust:1.95.0-bookworm`, `rustlang/rust:nightly-bookworm`, `python:3.12-slim-bookworm`, `debian:bookworm`
- Rust version is unchanged (1.95.0 / nightly); only the Debian suite changes

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -109,9 +109,11 @@ steps:
                 - bash
                 - -c
                 - |
-                  apt-get update && apt-get install -y pkg-config libssl-dev protobuf-compiler python3 python3-pip
-                  pip3 install -U pip setuptools wheel setuptools-rust build
-                  python3 -m build
+                  apt-get update && apt-get install -y pkg-config libssl-dev protobuf-compiler python3 python3-pip python3-dev python3-venv
+                  python3 -m venv /tmp/venv
+                  source /tmp/venv/bin/activate
+                  pip install -U pip setuptools wheel setuptools-rust build
+                  python -m build
         artifact_paths:
           - "dist/*.whl"
           - "dist/*.tar.gz"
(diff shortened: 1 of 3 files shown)


hsliuustc0106 (approved):
lgtm

PR #264, opened 2026-09-09
[CI]: Reclaim agent disk in premerge pipeline to fix ENOSPC

wuhang2014:
## Purpose

Premerge builds on the shared agents have started failing with `No space left on device` (e.g. [build 799](https://buildkite.com/vllm/router/builds/799)). The pipeline leaks disk on every build:

- **`Build Docker Image`** runs a multi-stage `docker build` whose `RUN cargo build --release` layer is multi-GB, tags it `vllm-router:<commit>`, and never pushes or deletes it. Nothing consumes the local tags downstream (release images are built and pushed by `release-pipeline.yml`), so each build permanently strands multi-GB of tagged images + build cache on `cpu_queue_premerge` agents. After ~800 builds the disks fill up.
- Cargo jobs (clippy, build, tests, wheel, python tests) each leave multi-GB `target/` dirs in the host-mounted checkout (`.:/workdir`). Intermediates are never reused across jobs (separate checkouts).
- The P/D GPU e2e job leaves a ~20GB `.venv` (vllm + torch cu130) plus `target/` in the `/workdir` checkout on `gpu_4_queue`, even when the job fails.

For reference, vllm-project/vllm CI avoids this class of problem by building images with buildx/bake into a registry with ECR layer cache, scheduling registry cleanup (`cleanup-nightly-builds.sh`), and running on ephemeral agents — none of which the router pipeline can assume from within the repo, so this PR makes the pipeline self-cleaning instead.

## Changes (all in `.buildkite/pipeline.yml`)

- **Docker Build**: prune stale `vllm-router:*` tags (except `latest`), dangling layers, and builder cache (`--keep-storage 10GB`) **before** building — this also reclaims disk on agents that are already full; after building, smoke-test the image with `vllm-router --version` inside `docker run` and `docker rmi` it. Local tags are unused downstream.
- **Cargo jobs** (clippy, unit/integration tests, wheel build, python tests): `rm -rf target` after the job.
- **Build Rust (Release)**: keep only `target/release/vllm-router` for the artifact upload; delete `deps`/`build`/`.fingerprint` intermediates.
- **P/D GPU job**: pre-clean `.venv`/`target` at start and `trap ... EXIT` so ~25GB is freed even when the job fails.

## Test Plan

- `python3` YAML parse of `.buildkite/pipeline.yml` — OK
- `bash -n` syntax check of every modified command block — OK
- pre-commit hooks (incl. `check-yaml`) — pass

## Test Result

Pipeline-level change only; behavior verified by YAML/bash syntax checks and the pre-commit hook. Effect will be observable on the next premerge runs: the Docker Build step now logs `df -h /` before/after pruning.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -53,6 +53,9 @@ steps:
                   apt-get update && apt-get install -y pkg-config libssl-dev protobuf-compiler
                   rustup component add clippy
                   cargo clippy --all-targets --all-features -- -D warnings
+                  # Free agent disk: target/ holds multi-GB of build artifacts
+                  # that are never reused across jobs (separate checkouts).
+                  rm -rf target || true
         agents:
           queue: "cpu_queue_premerge"
 
@@ -90,6 +93,9 @@ steps:
                 - |
                   apt-get update && apt-get install -y pkg-config libssl-dev protobuf-compiler
                   cargo build --release
+                  # Keep only the binary (uploaded as artifact below); the
+                  # intermediate artifacts are multi-GB and leak agent disk.
+                  rm -rf target/release/build target/release/deps target/release/.fingerprint target/release/incremental target/debug || true
         artifact_paths:
           - "target/release/vllm-router"
         agents:
@@ -114,6 +120,8 @@ steps:
                   source /tmp/venv/bin/activate
                   pip install -U pip setuptools wheel setuptools-rust build
                   python -m build
+                  # Free agent disk: cargo intermediates from the wheel build.
+                  rm -rf target || true
         artifact_paths:
           - "dist/*.whl"
           - "dist/*.tar.gz"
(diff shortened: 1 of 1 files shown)


hsliuustc0106 (changes requested):
[P1] Cleanup turns failed CI checks green (.buildkite/pipeline.yml:58; also lines 98, 124, 158, 179, and 205).

The six affected Docker plugin steps explicitly use bash -c without -e. Appending rm -rf ... || true replaces the build/test exit status with success, so failed Clippy checks, Rust builds, wheel builds, Rust unit/integration tests, and Python tests can pass CI. Please use an EXIT trap that preserves the original status, or explicitly capture and return the build/test status after cleanup.

I verified this against base 0519da5397f4 and PR head 0a19209517ee using the actual command blocks with stubbed build/test commands returning 42: all six base scripts returned 42, while all six PR scripts returned 0. This confirms the existing inline finding remains unresolved at the current head.

Validation: YAML parsing and bash syntax checks passed for all nine plugin scripts. Full Docker/GPU CI was not run.

PR #271, opened 2026-09-12
Preserve repeated response headers

### LINKED ISSUE #262: [Bug]: repeated Set-Cookie response headers are collapsed
I passed two response headers with the same name to `preserve_response_headers`:

```text
Set-Cookie: a=1
Set-Cookie: b=2
```

Expected: the returned `HeaderMap` contains both cookie fields.

Actual: it contains only `b=2`.

The property test shrank the failure to two repeated values. Tested on `main` commit `ab7e4537654caf8a2096bbe7c65e138754c72d6c` through the public header helper. I did not run a live HTTP proxy.

In [src/routers/header_utils.rs:24–36](https://github.com/vllm-project/router/blob/ab7e4537654caf8a2096bbe7c65e138754c72d6c/src/routers/header_utils.rs#L24-L36), the forwarding loop uses `HeaderMap::insert` for each value. Inserting the second value under the same name replaces the first one. The helper is called when forwarding HTTP responses, for example in [src/routers/http/router.rs:710–715](https://github.com/vllm-project/router/blob/ab7e4537654caf8a2096bbe7c65e138754c72d6c/src/routers/http/router.rs#L710-L715).

The test below appends two distinct `Set-Cookie` fields to the input and collects all output values. Expected: `["a=1", "b=2"]`. Actual: `["b=2"]`. Preserving separate values matters because each field represents a separate cookie. This verifies the shared helper, not a live upstream-to-client exchange.

### Standalone reproduction

In a checkout of `ab7e4537654caf8a2096bbe7c65e138754c72d6c`, save the code below as `tests/issue_262.rs`, then run:

```sh
cargo test --locked --test issue_262 -- --nocapture
```

<details>

```rust
use axum::http::{HeaderMap, HeaderValue};
use vllm_router_rs::routers::header_utils::preserve_response_headers;

#[test]
fn repeated_set_cookie_header_repro() {
    let mut input = HeaderMap::new();
    input.append("set-cookie", HeaderValue::from_static("a=1"));
    input.append("set-cookie", HeaderValue::from_static("b=2"));

    let output = preserve_response_headers(&input);
    let values = output
        .get_all("set-cookie")
        .iter()
        .map(|value| value.to_str().unwrap())
        .collect::<Vec<_>>();
    assert_eq!(values, ["a=1", "b=2"]);
}
```

</details>

### Verified failure output

```text
tests/issue_262.rs:16:5
  left: ["b=2"]
 right: ["a=1", "b=2"]
```

jiangwu25:
## Purpose

The router kept only the last value when a backend returned several headers with the same name. For example, two `Set-Cookie` headers for `a=1` and `b=2` became just `b=2`.

Change `preserve_response_headers` from `HeaderMap::insert` to `append`. Both cookies now reach the client in their original order.

Fixes #262.

## Test Plan

Add two tests: one checks both cookie values; the other checks that `Content-Type` survives while `Connection` and `Transfer-Encoding` are removed.

- `cargo test --locked --lib --bins`
- `cargo clippy --locked --all-targets --all-features -- -D warnings`

## Test Result

The cookie test failed before the fix. After the fix, all 486 library tests pass. Format and Clippy checks pass too.

### DIFF
--- a/src/routers/header_utils.rs
+++ b/src/routers/header_utils.rs
@@ -29,7 +29,7 @@ pub fn preserve_response_headers(reqwest_headers: &HeaderMap) -> HeaderMap {
         let name_str = name.as_str().to_lowercase();
         if should_forward_header(&name_str) {
             // The original name and value are already valid, so we can just clone them
-            headers.insert(name.clone(), value.clone());
+            headers.append(name.clone(), value.clone());
         }
     }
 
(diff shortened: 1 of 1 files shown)


wuhang2014 (approved):
LGTM

PR #272, opened 2026-09-12
Fix UTF-8 panic in partial stop checks

### LINKED ISSUE #259: [Bug]: Unicode partial stop sequences panic StopSequenceDecoder
I configured the hidden stop string `"éx"`, then processed one token that decodes to `"é"`.

Expected: `process_token` returns `Held` because the decoded text is a prefix of the configured stop string.

Actual: it panics:

```text
start byte index 1 is not a char boundary; it is inside 'é'
```

The property test also reproduces the panic with emoji. Tested on `main` commit `ab7e4537654caf8a2096bbe7c65e138754c72d6c` with an offline in-memory tokenizer.

In [src/tokenizer/stop.rs:191–205](https://github.com/vllm-project/router/blob/ab7e4537654caf8a2096bbe7c65e138754c72d6c/src/tokenizer/stop.rs#L191-L205), partial matching tries every byte length and slices `check_text` at `check_text.len() - i`. For `check_text = "é"`, the byte length is 2; the first iteration uses index 1, inside the UTF-8 encoding of `é`, and panics before testing the full prefix.

The included offline tokenizer maps token 3 to `é`. With hidden stop `éx`, processing that token should hold the prefix until the next token determines whether the stop matches. The reproduction exercises `process_token` directly and does not establish how a live server handles this panic.

### Standalone reproduction

In a checkout of `ab7e4537654caf8a2096bbe7c65e138754c72d6c`, save the code below as `tests/issue_259.rs`, then run:

```sh
cargo test --locked --test issue_259 -- --nocapture
```

<details>

```rust
use vllm_router_rs::tokenizer::stop::{SequenceDecoderOutput, StopSequenceConfig, StopSequenceDecoder};
use anyhow::Result;
use vllm_router_rs::tokenizer::traits::{Decoder, Encoder, Encoding, SpecialTokens, Tokenizer};

const PIECES: [&str; 8] = ["a", "b", " ", "é", "🙂", "<special>", "x", "\n"];

struct PieceTokenizer {
    special_tokens: SpecialTokens,
    piece_to_id: HashMap<&'static str, u32>,
}

impl PieceTokenizer {
    fn new() -> Self {
        let piece_to_id = PIECES
            .iter()
            .enumerate()
            .map(|(id, piece)| (*piece, id as u32))
            .collect();
        Self {
            special_tokens: SpecialTokens {
                bos_token: None,
                eos_token: Some("<special>".to_string()),
                unk_token: None,
                sep_token: None,
                pad_token: None,
                cls_token: None,
                mask_token: None,
                additional_special_tokens: Vec::new(),
            },
            piece_to_id,
        }
    }

    fn expected_decode(ids: &[u32], skip_special_tokens: bool) -> String {
        ids.iter()
            .filter_map(|id| PIECES.get(*id as usize))
            .filter(|piece| !(skip_special_tokens && **piece == "<special>"))
            .copied()
            .collect()
    }
}

impl Encoder for PieceTokenizer {
    fn encode(&self, input: &str) -> Result<Encoding> {
        let mut ids = Vec::new();
        let mut rest = input;
        while !rest.is_empty() {
            let Some(piece) = PIECES
                .iter()
                .filter(|piece| rest.starts_with(**piece))
                .max_by_key(|piece| piece.len())
            else {
                anyhow::bail!("input has no piece in the offline test vocabulary: {rest:?}");
            };
            ids.push(self.piece_to_id[piece]);
            rest = &rest[piece.len()..];
        }
        Ok(Encoding::Sp(ids))
    }

    fn encode_batch(&self, inputs: &[&str]) -> Result<Vec<Encoding>> {
        inputs.iter().map(|input| self.encode(input)).collect()
    }
}

impl Decoder for PieceTokenizer {
    fn decode(&self, token_ids: &[u32], skip_special_tokens: bool) -> Result<String> {
        Ok(Self::expected_decode(token_ids, skip_special_tokens))
    }
}

impl Tokenizer for PieceTokenizer {
    fn vocab_size(&self) -> usize {
        PIECES.len()
    }

    fn get_special_tokens(&self) -> &SpecialTokens {
        &self.special_tokens
    }

    fn token_to_id(&self, token: &str) -> Option<u32> {
        self.piece_to_id.get(token).copied()
    }

    fn id_to_token(&self, id: u32) -> Option<String> {
        PIECES.get(id as usize).map(|piece| (*piece).to_string())
    }
}

#[test]
fn known_failure_unicode_partial_stop_repro() {
    let tokenizer = Arc::new(PieceTokenizer::new());
    let config = StopSequenceConfig::default().with_stop_sequence("éx");
    let mut decoder = StopSequenceDecoder::new(tokenizer, config, false);

    assert_eq!(
        decoder.process_token(3).unwrap(),
        SequenceDecoderOutput::Held
    );
}
```

</details>

### Verified failure output

```text
src/tokenizer/stop.rs:201:41
start byte index 1 is not a char boundary; it is inside 'é' (bytes 0..2 of string)
```

jiangwu25:
## Purpose

Partial stop checks could panic on text such as `é` or `🙂`. The loop sliced text at each byte offset, which could fall inside a UTF-8 character.

Use `char_indices` to check suffixes only at valid character boundaries. The decoder still keeps the longest possible stop prefix until later text completes it or rules it out.

Fixes #259.

## Test Plan

Add four tests for Unicode and ASCII text, hidden and visible stops, stop completion, and text released after a mismatch. The tests use a small local tokenizer.

- `cargo test --locked --lib --bins`
- `cargo clippy --locked --all-targets --all-features -- -D warnings`

## Test Result

All four new tests hit the UTF-8 panic before the fix. After the fix, all 488 library tests pass. Format and Clippy checks pass too.

### DIFF
--- a/src/tokenizer/stop.rs
+++ b/src/tokenizer/stop.rs
@@ -196,11 +196,14 @@ impl StopSequenceDecoder {
             .iter()
             .chain(&self.config.visible_stop_sequences)
         {
-            // Check all possible suffixes that could be a prefix of stop_seq
-            for i in 1..=check_text.len().min(stop_seq.len() - 1) {
-                let suffix = &check_text[check_text.len() - i..];
+            // Check suffixes at UTF-8 character boundaries, shortest first.
+            for (start, _) in check_text.char_indices().rev() {
+                let suffix = &check_text[start..];
+                if suffix.len() >= stop_seq.len() {
+                    break;
+                }
                 if stop_seq.starts_with(suffix) {
-                    partial_match_len = partial_match_len.max(i);
+                    partial_match_len = partial_match_len.max(suffix.len());
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


wuhang2014 (approved):
LGTM

PR #280, opened 2026-09-16
fix: hold the concurrency slot for the whole streaming response

areslp:
## Purpose

`--max-concurrent-requests` does not bound concurrent streaming generations. It bounds how many requests may be *starting* at the same time.

`concurrency_limit_middleware` returns its token as soon as `next.run(request).await` returns:

```rust
if token_bucket.try_acquire(1.0).await.is_ok() {
    let response = next.run(request).await;
    // Return the token to the bucket
    token_bucket.return_tokens(1.0).await;
    response
}
```

`next.run()` returns once the upstream response **head** is ready. For `stream: true` the body is still being forwarded at that moment — `route_chat` spawns the forwarding task and hands back a `Body::from_stream` immediately (`src/routers/http/router.rs`, both the load-tracked and the untracked branch). So the slot is released at the first token instead of at the end of generation.

For an LLM workload that is the entire difference: time-to-first-token is milliseconds, a generation runs for seconds to minutes. With `--max-concurrent-requests 8` a fleet can end up with hundreds of streaming generations in flight, which is exactly the overload the flag is there to prevent.

Both existing tests in `py_test/integration/test_rate_limiting.py` use `"stream": False`, so the streaming path was never exercised.

## Approach

Move the token into the response body so it is returned once the body has been fully sent or dropped. A client disconnect drops the body, so a cancelled stream releases its slot as well.

`http-body-util` moves from `[dev-dependencies]` to `[dependencies]` so `BodyExt::map_frame` can be used in `src/`. It is already a transitive dependency of axum/hyper, so nothing new enters the build, and `map_frame` keeps trailers intact (unlike `into_data_stream`).

The mock worker's `--latency-ms` previously only applied to non-streaming responses; it now also applies between stream chunks, which is what lets a test hold a stream open.

## Test Plan

```bash
pip install -e .
pytest py_test/integration/test_rate_limiting.py -v -m integration
```

The new case `test_rate_limit_slot_held_for_streaming_response` starts one streaming request against a worker that streams for ~6s, then sends a second request 0.4s later with `max_concurrent_requests=1, queue_size=0`. The second request must be rejected with 429 while the first is still streaming.

## Test Result

Run on macOS (aarch64), rustc 1.98.1, Python 3.12.

**Before this change** (new tests applied on top of `main`, `src/` untouched) the new case fails and reproduces the bug:

```
py_test/integration/test_rate_limiting.py::test_rate_limit_and_queue PASSED
py_test/integration/test_rate_limiting.py::test_rate_limit_queue_and_timeout PASSED
py_test/integration/test_rate_limiting.py::test_rate_limit_slot_held_for_streaming_response FAILED

>       assert second == 429, f"slot was released mid-stream, got {second}"
E       AssertionError: slot was released mid-stream, got 200
E       assert 200 == 429

1 failed, 2 passed in 13.30s
```

The second request is admitted while the first is still streaming, which is the reported behaviour.

**After this change**:

```
py_test/integration/test_rate_limiting.py::test_rate_limit_and_queue PASSED       [ 33%]
py_test/integration/test_rate_limiting.py::test_rate_limit_queue_and_timeout PASSED [ 66%]
py_test/integration/test_rate_limiting.py::test_rate_limit_slot_held_for_streaming_response PASSED [100%]

3 passed in 16.92s
```

`cargo fmt --check` is clean. `cargo clippy --lib --bins` reports the same three pre-existing warnings as `main` (`src/server.rs:404`, `src/policies/consistent_hash.rs:51`, `src/core/token_bucket.rs:54`) and none in the changed code.


areslp:
The automated review is right, and it invalidates the framing of this PR. I reproduced it.

Holding the token for the body only reserves capacity for as long as the bucket does not hand the token back on its own — and it does. `TokenBucket::try_acquire` refills by elapsed time (`src/core/token_bucket.rs:57-63`) and the default refill rate *is* the concurrency limit (`src/server.rs:62`, `rate_limit_tokens_per_second.unwrap_or(max_concurrent_requests)`). So a slot reappears every `1 / max_concurrent_requests` seconds whether or not anything finished.

Probe on this branch, `max_concurrent_requests=1`, `rate_limit_tokens_per_second=1`, worker streaming for ~6s, second request sent after `gap`:

```
gap=0.4s  first_stream=200  second=429
gap=2.0s  first_stream=200  second=200   <- admitted, ~4s of the first stream still to run
```

So this change only bounds concurrency inside one refill interval. For real generations, which run far longer than that, it does not fix the overload the commit message claims to fix. And `test_rate_limit_slot_held_for_streaming_response` passes only because its gap is 0.4s — it masks exactly the case that matters, as the review says.

I should have caught this: I picked 0.4s precisely *because* the bucket refills at 1s, and treated that as a constraint on the test rather than as evidence that the mechanism under test is not a concurrency limit at all.

What still holds from this PR: the slot is currently released at the response head, so even within a refill interval the limit bounds time-to-first-token rather than generations, and the streaming path has no test coverage (both existing cases use `stream: false`). But that is a much smaller claim than the one I opened with.

The review's suggested direction looks correct to me — a non-refilling semaphore for concurrency, kept separate from request-rate limiting. That is a wider change than this PR, and it changes observable behaviour:

- `AppContext.rate_limiter` splits into a `Semaphore` for `max_concurrent_requests` and a `TokenBucket` for `rate_limit_tokens_per_second`
- `QueueProcessor` waits on the semaphore instead of `acquire_timeout`
- `rate_limit_tokens_per_second` left unset would mean "bound concurrency, do not rate limit", instead of today's implicit rate limit
- requests that today trickle through on refill would queue or get a 429

That is a design call for maintainers rather than something to slip into this PR, so I would rather ask than guess:

1. Should I rework this PR into the semaphore split?
2. Or keep this one scoped to the head-vs-body release plus the missing streaming coverage, with the semaphore as a separate PR?
3. Or would you rather own the concurrency redesign, in which case I will close this and open an issue with the reproduction above?

Happy to convert this to a draft in the meantime so it is not mistaken for a complete fix. #281 (`Retry-After` on the rate-limit 429) is independent of all of this and unaffected.

Issue #285, opened 2026-09-18
[Feature]: Program-level Progress-TTL scheduling in vLLM Router


warriorsniu:
### Agent workload characteristics

An agent workload is a long-lived execution composed of alternating LLM calls and tool calls. Each LLM request is therefore one step of a larger workflow rather than an independent unit of work. The next request commonly contains most of the preceding context plus the latest model output and tool result, so adjacent requests from the same agent have substantial prefix overlap and a high potential cache-hit rate.

Tool calls are often short relative to the cost of reconstructing a long context. This creates strong temporal continuity: after one LLM request completes, the same agent may return to the serving system before its KV cache would need to be displaced for useful work. In one representative 32-task/16-concurrency coding-agent trace, 86.46% of adjacent same-agent requests returned within one second and 95.12% returned within ten seconds. Preserving that context can avoid cold prefill and its interference with concurrent decode work.

![Distribution of adjacent same-agent request intervals](https://raw.githubusercontent.com/warriorsniu/router/docs/program-level-progress-ttl-rfc-assets/docs/rfc-assets/program-level-progress-ttl/tool_call_interval_distribution.png)

Agent frameworks also introduce concurrency that ordinary chat traffic does not capture. A lead agent can fan out into multiple subagents, while each agent independently alternates between reasoning and tool execution. Their contexts grow over successive rounds and coexist in the KV pool. The combination of long reusable prefixes, short return intervals, and multi-agent fan-out creates both a large opportunity for KV reuse and sustained pressure on finite KV capacity.

![Incremental context growth and potential prefix-cache reuse](https://raw.githubusercontent.com/warriorsniu/router/docs/program-level-progress-ttl-rfc-assets/docs/rfc-assets/program-level-progress-ttl/workload_context_cache_reuse.png)

### Cache thrashing under concurrent agent workloads

When admission is not constrained by KV capacity, request-level FCFS scheduling can interleave many agent contexts. The intervening allocations push reusable context toward the LRU eviction boundary. As a result, an agent can return from a short tool call after only a brief request interval and still find that most of its KV cache has already been evicted. The next request must then perform a cold prefill, increasing its own latency, interfering with concurrent decode work, and reducing system throughput.

![Native vLLM throughput and cache-hit degradation under increasing agent load](https://raw.githubusercontent.com/warriorsniu/router/docs/program-level-progress-ttl-rfc-assets/docs/rfc-assets/program-level-progress-ttl/native_vllm_peak_task_scaling.png)

[...]

Issue #244, opened 2026-09-08
[WIP][Help Wanted] vLLM Router 2026 Roadmap


fake0fan:
# vLLM Router 2026 Roadmap Issues

> Planning date: September 8, 2026  
> Code baseline: vLLM Router `main@1d10e71` (identical to upstream `main` at the time of review)  
> Objective: define an actionable roadmap for the remainder of 2026 around vLLM-native integration, lightweight high-performance routing, exact KV awareness, and agent-aware scheduling.  
> Scope: CPU fast path, exact KV routing, agent-aware scheduling, E/P/D, and Router HA.

## Overview

| # | Roadmap issue | Current foundation | Priority | 2026 target | External value |
|---:|---|---|---|---|---|
| 1 | Integrate and Optimize the CPU-side Inference Fast Path | ◐ Rust tokenizer, streaming utilities, and benchmarks exist, but they do not yet form an end-to-end request fast path | P0 | Tokenizer cache, request-scoped reuse, and JSON/streaming hot-path optimization | Demonstrate the performance advantage of a lightweight Rust router |
| 2 | Event-driven Exact KV-aware Routing | ○ Exact KV awareness is missing; the current cache-aware policy uses an approximate request-history tree | P0 | KV Events, an exact index, unified cost, and safe fallback | Flagship capability for the year |
| 3 | Extend Session Affinity to Agent/Program-aware Scheduling | ◐ Session-ID-based consistent hashing exists | P1 | Program metadata, a program-aware policy, and a KV TTL prototype | Differentiate for agentic workloads |
| 4 | Extend P/D Routing to E/P/D | ◐ P/D, DP-rank routing, and KV connectors exist; the Encode stage does not | P1 | Encode pool and three-stage routing MVP | Extend multimodal scheduling |
| 5 | Complete Multi-Router HA and State Recovery | ◐ Health checks, retries, readiness, and process-level graceful shutdown exist; multi-Router state HA does not | P0 foundation / P1 complete | Drain and recovery foundations; active-active as a stretch goal | Make stateful routing production-ready |

## Roadmap principles

- Do not turn vLLM Router into a broad, all-in-one gateway; its core remains vLLM inference routing.
- Prioritize CPU optimizations that directly benefit tokenization, routing, and the streaming fast path.
- Make exact KV state trustworthy before optimizing policy gains.
- Agent hints must be optional, and ordinary OpenAI-compatible requests must remain compatible.
[...]

PR #286, opened 2026-09-18
[Feature] Add Program-level Progress-TTL scheduling

warriorsniu:
## Purpose

Implement the Program-aware scheduling scope tracked by [Roadmap issue #244](https://github.com/vllm-project/router/issues/244) and specified in [RFC #285](https://github.com/vllm-project/router/issues/285). The feature treats a continuous `LLM -> tool call -> LLM` interaction as one Program, retains requests before forwarding, and controls when and where paused Programs resume so short tool calls are less likely to lose reusable KV cache under pressure.

This is an opt-in Router feature. When Program scheduling is disabled, or when a request carries no supported Program identity, the native routing, retry, forwarding, response, and metrics paths remain unchanged.

The PR delivers the complete feature in one reviewable change because identity, RequestPool retention, Progress-TTL decisions, backend observations, HTTP integration, and diagnostics share one state machine and contract. Benchmark Replay changes are intentionally outside this PR.

### Program request contract

- Normalize identity from `vllm_xargs.agentic_context`, known agent hints, and supported framework headers with an explicit precedence order.
- Represent task, Program, parent, blocking, expected-resume, priority, deadline, expected-output-length, and KV-retention hints with typed request-scoped fields.
- Preserve identity-free fallback and compatibility behavior for requests that do not provide Program metadata.
- Resolve parent/root relationships without making optional upstream hints authoritative lifecycle facts.

### Scheduling runtime

- Add a Program registry and RequestPool that retain, release, cancel, and clean up request generations safely.
- Model Program scheduling state independently from request execution status.
- Maintain per-Rank rolling factors, backend capacity observations, token estimates, Program-private occupancy, and observation-epoch accounting.
- Support initial Program binding through Consistent Hash, Program Round Robin, Least Program Count, Reasoning Token Balance, and Cache-aware binding policies.

### Progress-TTL policy

- Admit new Programs and resume paused Programs only when the selected Rank has sufficient current capacity; future-pause relief is reserved for capacity repair and is not treated as immediately free capacity.
- Protect short acting intervals with per-request TTL decisions derived from observed continuity and estimated cache-miss impact, bounded by the explicit `max_acting_ttl_seconds` configuration.
- Make the cold-prefill cost model and batch decode-throughput surface configurable so deployments can use coefficients calibrated for their model, hardware, parallel layout, and inference-engine version.
- Reserve bounded context growth for active Programs and allow batch-gain admission only when the modeled decode benefit exceeds continuity loss.
- Pause Programs for TTL expiration, capacity repair, or maximum continuous service, while keeping those causes explicit for later placement decisions.
- Do not reclaim active acting Programs from admission or resume; acting TTL and terminal lifecycle own that transition.
- Bound starvation with force-resume deadlines, request priority, and a one-shot lower resume priority after maximum continuous service.
- Default Rank-local resume order to MRU while retaining FCFS as an option. Global Queue attempts all feasible local resumes before cross-Rank placement; cross-Rank candidates use fixed FCFS ordering after force-resume, request-priority, and one-shot-yield tiers.

### Router integration

- Integrate exactly four scheduling interaction classes: request arrival, streaming output observation, request completion, and periodic backend polling.
- Treat every Router retry as a fresh arrival while completing or cancelling each dispatched attempt exactly once.
- Parse OpenAI- and Anthropic-compatible usage without changing backend response bytes, then feed authoritative prompt, cached-prompt, and completion counts back into token estimation.
- Keep initial Program binding separate from runtime resume placement so existing request-level load-balancing policies are not silently repurposed.
- Expose authenticated scheduling diagnostics and bounded Prometheus metrics without constructing Program snapshots on the normal forwarding hot path.

### Global Queue and engine-control boundary

Global Queue may place a paused Program on another Rank only after Rank-local opportunities and the configured capacity/headroom gates have been evaluated. The Router controls request admission and placement, but this PR does not claim that an inference engine executed physical KV eviction, offload, or prefetch. A future Router-to-engine KV command must receive explicit accepted/rejected/unsupported feedback and terminal applied/failed/expired feedback before Router accounting changes.

### Files added or modified

The implementation adds one scheduling module and a narrow HTTP adapter while extending existing Router configuration, policy, protocol, metrics, lifecycle, and test entry points. `added` and `modified` are relative to the current vLLM Router `main` branch.

```text
router/
├── README.md                                      # modified — link Program scheduling configuration
├── docs/
│   └── program_scheduling.md                     # added — complete configuration reference, identity contract and model calibration
├── benches/
│   └── request_processing.rs                     # modified — keep request-processing benchmarks compatible
├── py_src/vllm_router/
│   ├── router.py                                 # modified — expose Program scheduling in the Python entry point
│   └── router_args.py                            # modified — parse Program scheduling configuration
├── src/
│   ├── config/
│   │   ├── types.rs                              # modified — define Program scheduling and calibration settings
│   │   └── validation.rs                         # modified — validate capacity, timing and model coefficients
│   ├── core/
│   │   └── worker_registry.rs                    # modified — provide revisioned immutable target snapshots
│   ├── policies/
│   │   ├── cache_aware.rs                        # modified — expose reusable prefix-affinity scoring
│   │   └── mod.rs                                # modified — export policy primitives used by Program binding
│   ├── program_scheduling/
│   │   ├── binding/mod.rs                        # added — initial Program-to-Rank binding policies
│   │   ├── contracts.rs                          # added — scheduler errors and decision contracts
│   │   ├── diagnostics.rs                        # added — bounded on-demand state snapshots
│   │   ├── domain.rs                             # added — Program state, status, token and dispatch types
│   │   ├── factors.rs                            # added — rolling workload and decision factors
│   │   ├── identity.rs                           # added — canonical and compatibility identity parsing
│   │   ├── lineage.rs                            # added — parent/root Program relationship tracking
│   │   ├── mod.rs                                # added — module boundary and public exports
│   │   ├── observations.rs                       # added — normalized per-Rank backend observations
│   │   ├── policy.rs                             # added — Progress-TTL, capacity and batch-gain calculations
│   │   ├── request_pool.rs                       # added — request retention, wakeup and cancellation
│   │   ├── runtime.rs                            # added — generation-safe Program registry
│   │   ├── scheduler.rs                          # added — orchestration and observation epochs
│   │   ├── scheduler_admission.rs                # added — admission, resume ordering and capacity checks
│   │   ├── scheduler_fairness.rs                 # added — force resume and bounded-service fairness
│   │   ├── scheduler_global.rs                   # added — local-first and cross-Rank Global Queue placement
│   │   ├── scheduler_lifecycle.rs                # added — completion, TTL, pause and release transitions
│   │   └── scheduler_state.rs                    # added — per-Rank ledgers, queues and target snapshots
│   ├── protocols/
│   │   └── spec.rs                               # modified — retain structured content and identity fields
│   ├── routers/http/
│   │   ├── mod.rs                                # modified — export the Program adapter
│   │   ├── program_adapter.rs                    # added — connect HTTP events to ProgramScheduler
│   │   └── router.rs                             # modified — gate forwarding through RequestPool admission
│   ├── routers/mod.rs                            # modified — export Router integration types
│   ├── lib.rs                                    # modified — expose configuration and scheduling APIs
│   ├── main.rs                                   # modified — construct Program scheduling from CLI configuration
│   ├── metrics.rs                                # modified — add bounded Program and transition metrics
│   ├── server.rs                                 # modified — manage periodic observation and scheduling work
│   ├── token_estimator.rs                        # added — feedback-calibrated request token estimation
│   └── tree.rs                                   # modified — share prefix-tree scoring with Program binding
├── py_test/unit/
│   └── test_arg_parser.py                        # modified — validate Python configuration plumbing
└── tests/
    ├── api_endpoints_test.rs                     # modified — cover opt-in, fallback and HTTP lifecycle behavior
    ├── test_dp_routing.rs                        # modified — cover concrete DP-Rank integration
    └── test_pd_routing.rs                        # modified — preserve P/D routing compatibility
```

## Test Plan

```bash
cargo fmt --check
cargo test program_scheduling:: --lib
cargo test routers::http::program_adapter --lib
cargo test token_estimator --lib
cargo test test_program_request_is_tracked_without_changing_response --test api_endpoints_test -- --test-threads=1
cargo test --lib --tests -- --skip test_batch_streaming --test-threads=1
cargo check --all-targets
git diff --check
```

The performance plan replays the same task list, concurrency, sampling configuration, backend capacity, and generated token plan against native request routing and Program scheduling. It checks task/request completion, exact token totals, prompt-token cache hit, throughput, request latency, task completion time, Rank placement, pause reason, TTL, capacity ledger, and queue contents.

## Test Result

The current implementation passes `cargo fmt --check` and all 64 `program_scheduling` library tests. These tests cover identity precedence and fallback, Program state transitions, request retention and cancellation, lifecycle generation safety, admission/resume capacity, capacity repair, acting Program non-reclamation, TTL behavior, force resume, FCFS/MRU Rank-local ordering, fixed-FCFS cross-Rank ordering, Global Queue local-first behavior, binding policies, rolling factors, diagnostics, and deterministic tie-breaks. Configuration tests additionally cover complete and partial calibration-model overrides plus invalid coefficient rejection.

The primary DP2/TP1 comparison used `Qwen3-Coder-30B-A3B-Instruct-FP8`, native vLLM `AsyncScheduler`, 266,864 KV tokens per Rank, 24 tasks at concurrency 24, and the same 744-request replay plan. Every arm completed 24/24 tasks and 744/744 requests with exactly 15,038,326 input tokens and 113,849 output tokens. Prompt hit is computed as cached prompt tokens divided by prompt tokens, not cache-query hit rate. The Program scheduling arm used the default MRU Rank-local resume order with Least Program Count binding.

| Method | Prompt hit | Wall time | Mean task time | Mean / P99 request latency | Output throughput |
|---|---:|---:|---:|---:|---:|
| Consistent Hash baseline | 52.74% | 1,926.1s | 1,836.1s | 59.57s / 309.49s | 59.11 tok/s |
| Request-level Cache-aware baseline | 63.74% | 1,577.0s | 1,505.2s | 48.73s / 299.91s | 72.19 tok/s |
| Program Scheduler | **76.32%** | **1,164.5s** | **1,066.3s** | **34.43s / 206.97s** | **97.77 tok/s** |

| Program Scheduler improvement | Versus Consistent Hash | Versus request-level Cache-aware |
|---|---:|---:|
| Prompt hit | +23.58 percentage points | +12.58 percentage points |
| Wall time | -39.5% | -26.2% |
| Mean task completion time | -41.9% | -29.2% |
| Mean request latency | -42.2% | -29.3% |
| P99 request latency | -33.1% | -31.0% |
| Output throughput | +65.4% | +35.4% |

The baseline comparison demonstrates the workload-level benefit of Program scheduling under cache pressure. `cargo check --all-targets` and `git diff --check` pass on the current working tree. The full Router library/integration run completed with 588 passing tests; 8 WASM middleware tests could not run because the configured rustup mirror returned 404 while installing the unrelated `wasm32-wasip2` standard-library component.

---
<details>

- [x] The purpose of the PR, including Roadmap issue #244 and RFC #285.
- [x] The test plan, including targeted, integration, formatting, and compile checks.
- [x] The current test result and recorded end-to-end comparison.

</details>


hsliuustc0106 (commented):
Risk-focused review pass at `7a91c7f7` (this is not a full line-by-line audit of ~9.6k new lines — flagging that upfront given the size and the single-state-machine design).

What I verified as solid:
- **Default-off gating**: `program_scheduling` is `Option<ProgramSchedulingConfig>` plus a per-request enable key; when disabled or the request carries no identity, the native routing/retry/forwarding path is untouched (early `None`). Matches the PR's compatibility claim.
- **Retention is bounded**: parked requests wait with a queue timeout and `ScheduleError::QueueTimeout` maps to 429 — no unbounded parking.
- **No lost wakeups**: each retained request owns a fresh `Notify` and `notify_front` uses `notify_one`, which stores a permit for a not-yet-arrived waiter.
- **Disconnect while parked**: `AdmissionGuard` cancels the retained entry on Drop and disarms after admission, so a parked request can't outlive its HTTP task.
- **Memory bounds**: SSE usage accumulation caps the partial buffer at 1 MiB; completion is swap-once (`finished` flag) with Drop → `complete(false)`.

Findings:

**P3 — forced-target unavailability returns a misleading 503 "no_available_workers"** (`src/routers/http/router.rs`, retry closure): when the scheduler binds a Program to a specific worker and that worker is momentarily unavailable (`get_by_url(target).filter(is_available)`), the request fails with the same error used for "no workers at all". Defensible for KV affinity, but a distinct error/message would help operators tell "bound target busy" from "pool empty".

Scope notes worth a maintainer look:
1. `acquire_program_completion` runs *inside* the retry closure, so a retryable failure re-acquires for the same Program. I did not trace whether re-acquire after an already-admitted dispatch is idempotent or double-counts Program occupancy/active-program caps.
2. The admission/fairness/observation-epoch state machine is the heart of the feature; this pass checked its boundaries (gating, bounds, lifecycle), not its internal logic.

---
*From an automated daily review pass over new/updated PRs (head SHA frozen at `7a91c7f7`).*

src/routers/http/router.rs
```diff
@@ -167,6 +184,43 @@ impl Router {
...
+            let provider = VllmMetricsObservationProvider::new(
+                ctx.client.clone(),
+                ctx.router_config.api_key.clone(),
+            );
+            Arc::new(tokio::spawn(async move {
+                let mut interval = tokio::time::interval(scheduler.metrics_interval());
+                interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Skip);
+                loop {
+                    interval.tick().await;
+                    let targets = scheduler.all_targets();
```
david6666666, line 221: [P1] This loop only sees the scheduler's last target snapshot. Worker add/remove/health changes bump the registry revision, but nothing here refreshes the scheduler, so an idle removal can leave a dead target in `all_targets()` and its capacity ledger. Could this sync from `worker_registry` (or subscribe to the revision) before polling?

src/program_scheduling/identity.rs
```diff
@@ -0,0 +1,943 @@
...
+            value
+                .to_str()
+                .map(|value| value.trim().to_string())
+                .map_err(|_| ScheduleError::InvalidIdentity(format!("Invalid {name} header")))
+        })
+        .transpose()
+        .map(|value| value.filter(|value| !value.is_empty()))
+}
+
+fn context_string(
```
david6666666, line 664: [P1] `build()` caps only `program_id`/`placement_key`/`model_pool`, but this helper returns unbounded strings for parent/root/agent/request metadata that are retained in Program state. A caller can therefore create a much larger retained identity than the intended 256-byte limit. Please apply the same bound to every persisted string field.

src/program_scheduling/scheduler.rs
```diff
@@ -0,0 +1,762 @@
...
+    /// Install raw observations and preserve transitions committed during I/O.
+    pub fn apply_observations(
+        &self,
+        epoch: BackendObservationEpoch,
+        observations: impl IntoIterator<Item = BackendObservation>,
+    ) {
+        let mut state = self.state.lock();
+        for raw in observations {
+            let target_id = raw.target_id.clone();
+            let Some(checkpoint) = epoch.checkpoints.get(&raw.target_id) else {
```
david6666666, line 335: [P1] The epoch only proves the target existed when polling started. If `sync_target_snapshot` removes or rebinds it while the request is in flight, this path can recreate an observation for that target and resurrect stale capacity when the same URL is added again. Please drop observations for targets no longer in the current snapshot, or fence them by target revision.

src/routers/http/router.rs
warriorsniu: Thanks, this was a real gap. Fixed in [`e782bbe`](https://github.com/vllm-project/router/commit/e782bbed589255ad7d0d0ee8e672b879ea27849b): before each polling pass, the observation task now collects available targets separately for every tracked model pool and verifies that `WorkerRegistry::revision()` remained stable during collection. A mixed view is discarded and collected again; a stable view goes through the existing `sync_target_snapshot` path, which owns target cleanup, placement invalidation, and Program rebinding. The regression test covers idle health/removal updates and model-pool isolation without requiring another request arrival.

src/program_scheduling/identity.rs
warriorsniu: Agreed, fixed in [`e782bbe`](https://github.com/vllm-project/router/commit/e782bbed589255ad7d0d0ee8e672b879ea27849b). A final validation pass now runs after fallback and derived-ID construction, so every persisted identity and request-hint string uses the 256-byte component limit while the placement hash key keeps its separate 512-byte limit. Tests cover canonical context, framework headers, `agent_hint`, derived IDs, request metadata, the accepted boundary, and UTF-8 byte length.

src/program_scheduling/scheduler.rs
warriorsniu: Agreed, fixed in [`e782bbe`](https://github.com/vllm-project/router/commit/e782bbed589255ad7d0d0ee8e672b879ea27849b). This remains separate from the pre-poll WorkerRegistry refresh: the scheduler now captures a monotonic target-snapshot revision in `BackendObservationEpoch`, rejects the whole batch if topology changed during metrics I/O, and checks current membership before insertion. Tests cover removal and same-URL remove/re-add, and a rejected batch does not consume its ledger checkpoint.

warriorsniu:
> Risk-focused review pass at `7a91c7f7` (this is not a full line-by-line audit of ~9.6k new lines — flagging that upfront given the size and the single-state-machine design).
> 
> What I verified as solid:
> 
> * **Default-off gating**: `program_scheduling` is `Option<ProgramSchedulingConfig>` plus a per-request enable key; when disabled or the request carries no identity, the native routing/retry/forwarding path is untouched (early `None`). Matches the PR's compatibility claim.
> * **Retention is bounded**: parked requests wait with a queue timeout and `ScheduleError::QueueTimeout` maps to 429 — no unbounded parking.
> * **No lost wakeups**: each retained request owns a fresh `Notify` and `notify_front` uses `notify_one`, which stores a permit for a not-yet-arrived waiter.
> * **Disconnect while parked**: `AdmissionGuard` cancels the retained entry on Drop and disarms after admission, so a parked request can't outlive its HTTP task.
> * **Memory bounds**: SSE usage accumulation caps the partial buffer at 1 MiB; completion is swap-once (`finished` flag) with Drop → `complete(false)`.
> 
> Findings:
> 
> **P3 — forced-target unavailability returns a misleading 503 "no_available_workers"** (`src/routers/http/router.rs`, retry closure): when the scheduler binds a Program to a specific worker and that worker is momentarily unavailable (`get_by_url(target).filter(is_available)`), the request fails with the same error used for "no workers at all". Defensible for KV affinity, but a distinct error/message would help operators tell "bound target busy" from "pool empty".
> 
> Scope notes worth a maintainer look:
> 
> 1. `acquire_program_completion` runs _inside_ the retry closure, so a retryable failure re-acquires for the same Program. I did not trace whether re-acquire after an already-admitted dispatch is idempotent or double-counts Program occupancy/active-program caps.
> 2. The admission/fairness/observation-epoch state machine is the heart of the feature; this pass checked its boundaries (gating, bounds, lifecycle), not its internal logic.
> 
> _From an automated daily review pass over new/updated PRs (head SHA frozen at `7a91c7f7`)._

Thanks, the "no_available_workers" issue is  fixed in [`e782bbe`](https://github.com/vllm-project/router/commit/e782bbed589255ad7d0d0ee8e672b879ea27849b). The forced Program-target miss now keeps the retryable 503 status but uses the bounded `program_target_unavailable` metric reason and `Program-bound worker is unavailable` response without exposing the internal URL. The existing `no_available_workers` path remains specific to an empty or fully unavailable worker pool.

The "acquire_program_completion" issue is fixed: each backend retry is modeled as a fresh request arrival after the preceding dispatch completes. I kept acquisition inside the retry closure and documented that contract in [`e782bbe`](https://github.com/vllm-project/router/commit/e782bbed589255ad7d0d0ee8e672b879ea27849b). The new regression test runs a failed attempt followed by a successful attempt, exercises repeated explicit finish plus Drop, and verifies that the final Program has no in-flight or retained request leak.

src/program_scheduling/identity.rs
```diff
@@ -0,0 +1,1059 @@
...
+                    "vllm_xargs.agentic_context must contain a bounded JSON object".to_string(),
+                ));
+            }
+            serde_json::from_str::<JsonValue>(value).map_err(|_| {
+                ScheduleError::InvalidIdentity(
+                    "vllm_xargs.agentic_context must contain a bounded JSON object".to_string(),
+                )
+            })?
+        }
+        value => value.clone(),
```
david6666666: [P1] The 16 KiB cap still only applies to the string form. The object form is cloned here without a serialized-size check, so a large transparent request can carry a much larger `agentic_context` and incur multiple JSON copies before field validation. Can we enforce the same bound for both forms?

src/routers/http/router.rs
```diff
@@ -181,9 +237,194 @@ impl Router {
...
+                        model => worker_registry.get_by_model_fast(model),
+                    };
+                    let targets: Arc<[ProgramTarget]> = Self::program_targets(&workers).into();
+                    (model_pool.clone(), targets)
+                })
+                .collect::<Vec<_>>();
+            if worker_registry.revision() != revision {
+                continue;
+            }
+            for (model_pool, targets) in snapshots {
```
david6666666: [P2] The registry revision is checked before these snapshots are installed. A worker can be removed immediately after the check, leaving the scheduler with a target that is already absent from the registry until the next polling tick. Could we re-check after syncing or expose an atomic snapshot-plus-revision?

src/routers/http/router.rs
```diff
@@ -181,9 +237,194 @@ impl Router {
...
+            return;
+        }
+        loop {
+            let revision = worker_registry.revision();
+            let snapshots = model_pools
+                .iter()
+                .map(|model_pool| {
+                    let workers = match model_pool.as_str() {
+                        "default" => worker_registry.get_all(),
+                        "unknown" => Vec::new(),
```
david6666666: [P2] `unknown` is also the Worker trait fallback model ID when no model label is present. Treating it as an always-empty sentinel drops real workers for a request whose model is literally `unknown`; a separate unresolved-model marker would avoid that collision.

src/routers/http/program_adapter.rs
```diff
@@ -0,0 +1,293 @@
...
+        accumulator.observe_json(
+            br#"{"usage":{"input_tokens":10,"cache_read_input_tokens":30,"cache_creation_input_tokens":20,"output_tokens":5}}"#,
+        );
+        assert_eq!(accumulator.observation.prompt_tokens, Some(60));
+        assert_eq!(accumulator.observation.cached_prompt_tokens, Some(30));
+        assert_eq!(accumulator.observation.completion_tokens, Some(5));
+    }
+
+    #[tokio::test]
+    async fn retry_attempts_complete_exactly_once_without_occupancy_leak() {
```
david6666666, line 249: [P2] This test calls `finish()` manually but never drives `RetryExecutor` or a retryable backend response, so it does not cover the route-level retry path this regression concerns. Please add an integration case that retries a failed attempt and checks occupancy across both attempts.

src/program_scheduling/identity.rs
warriorsniu: Thanks. Fixed in [`8cee321`](https://github.com/warriorsniu/router/commit/8cee3216bdc76033478f64c61f181cc525723dcd). Both string and object forms of `vllm_xargs.agentic_context` are now subject to the 16 KiB total-size limit before the object is cloned. The object path uses `serde_json::to_writer` with a bounded writer, so rejecting an oversized context does not allocate another input-sized buffer. The per-field persisted-string limits remain as a separate check. Tests cover accepted and rejected string/object forms, including an oversized object made only of unknown fields.

src/routers/http/router.rs
warriorsniu: Thanks. Fixed in [`605f12c`](https://github.com/warriorsniu/router/commit/605f12c537fa272c7ade57f0e6a591942cc25fb0). All tracked model-pool snapshots are now installed through one bulk scheduler update under a single state lock. The refresh loop checks `WorkerRegistry::revision()` both before and after that installation and immediately recollects/reinstalls if the revision changed. I added a deterministic regression test that removes a worker across the install boundary and verifies that the final scheduler view contains no stale target. The route-level live-registry check and the scheduler observation-epoch fence remain in place as the dispatch and metrics-I/O safeguards.

src/routers/http/router.rs
warriorsniu: Thanks. Fixed in [`7586db1`](https://github.com/warriorsniu/router/commit/7586db125d06578bc86ff0e20dacf791002e6bfa). I separated the Program scheduler pool identity from the WorkerRegistry lookup key. Named models, including the literal model name `unknown`, now use a tagged `program:model:<name>` pool. Unresolved request models use the distinct `program:fallback:unlabeled` pool while still looking up unlabeled workers through the registry's `unknown` fallback key. The target cache is keyed by the scheduler pool, so these cases cannot collide. The added tests cover a literal `unknown` model, unresolved models sharing the unlabeled fallback, and observation refresh for tagged model pools.

src/routers/http/program_adapter.rs
warriorsniu: Thanks. Added in [`b3d084a`](https://github.com/warriorsniu/router/commit/b3d084a9228724cc50eb53b38ebedf5a58762696). The new HTTP integration test drives the full Router retry path against a deterministic backend that returns 503 on the first generation request and 200 on the second. It verifies two backend attempts, a successful final response, a single Program generation, and zero retained/in-flight/waiting requests and Router occupancy after the retry completes. The focused `ProgramCompletion::finish()` idempotency test remains in place as unit-level coverage.

david6666666 (approved):
Re-reviewed at a77107f. The earlier findings are addressed: both context forms are size-bounded, persisted identity fields are capped, target refresh and observation fencing are in place, model fallback keys are separated, and the new HTTP test exercises a real 503-to-200 retry with Program accounting checks.

The calibration warnings and observability changes also look reasonable. No remaining blockers found in this follow-up review.

Validation: git diff --check passes. I could not rerun the Rust tests locally because no Rust toolchain is installed; the test results in the PR description are author-reported.

src/routers/http/router.rs
```diff
@@ -1730,6 +2154,19 @@ impl RouterTrait for Router {
             request_builder = request_builder.json(&body);
         }
 
+        if let Some(headers) = headers {
```
wuhang2014: **[blocker]** This header-forwarding loop runs for every transparent-proxy request, including when Program scheduling is disabled — on `main` this path sent only the JSON body plus router auth, so native forwarding behavior changes on upgrade.

Two concrete problems:
- The copied headers include the client's `Authorization`; the router's own `Authorization: Bearer <api_key>` is appended a few lines below, producing **two Authorization header values** whenever `--api-key` is set. Per RFC 7235 a request must carry exactly one; backends may reject it or honor the wrong credential.
- There is no hop-by-hop denylist, so `Connection`, `Keep-Alive`, `Upgrade`, `TE`, `Proxy-Authorization`, etc. are forwarded to the backend (RFC 7230 §6.1 says these must not be forwarded).

Suggested fix: gate the loop on `program_completion.is_some()` so the native path is byte-identical when the feature is off, and when copying, skip `AUTHORIZATION` plus the standard hop-by-hop names (an allowlist or a helper mirroring `header_utils::preserve_response_headers` would match repo conventions).

src/protocols/spec.rs
```diff
@@ -543,19 +557,118 @@ impl GenerationRequest for ChatCompletionRequest {
     }
 
     fn extract_text_for_routing(&self) -> String {
```
wuhang2014, line 559: **[blocker]** This rewrite of `extract_text_for_routing()` changes native routing for **all** deployments, even with `program_scheduling: None`:

- **consistent_hash / rendezvous_hash regression for `session_params` users**: previously this returned `session_params.session_id`, so `hash_key::extract_hash_key` produced the stable key `request:<session_id>` for every turn of a session. Now it returns the whole role-prefixed conversation, which is not JSON, so extraction falls through to the `request_hash:<fbi(text)>` fallback — and the hash changes on **every turn** as messages accumulate, breaking session affinity (and the KV-cache locality it provides).
- **cache_aware (the default policy)**: the prefix tree previously received session-id strings (or nothing); it now receives full conversation text for every request — different affinity semantics plus per-request tree insert cost/growth.

The change is clearly deliberate (needed for cache-aware Program binding), but it should not alter the disabled-feature path. Suggestion: keep the session-id-first behavior for the native `select_worker_for_model` text and use the conversation-text extraction only for the Program identity/binding path (e.g. a separate trait method like `extract_program_identity_payload`'s sibling), or gate on `program_scheduler.is_some()` + `binding_strategy == cache_aware`. If maintainers do want the new semantics for everyone, that should be called out explicitly in the PR description as a breaking routing change, not shipped inside an opt-in feature.

src/routers/http/program_adapter.rs
```diff
@@ -0,0 +1,293 @@
...
+
+    pub(super) fn dispatch(&self) -> &ProgramDispatch {
+        &self.inner.dispatch
+    }
+
+    pub(super) fn observe_json(&self, body: &[u8]) {
+        self.inner.usage.lock().observe_json(body);
+    }
+
+    pub(super) fn observe_sse_chunk(&self, chunk: &[u8]) {
```
wuhang2014, line 58: **[warning]** Every `data:` SSE line of every tracked stream is fully parsed with `serde_json::from_str` even though only the final chunk(s) of a response carry `usage` — for long agent generations this is thousands of full JSON parses (plus a `drain(..)`→`Vec` allocation per line) per request on the forwarding hot path.

Cheap fix: before parsing, check the line for the byte sequence `usage` (e.g. `memchr`/`windows::contains`); skip lines without it. That eliminates ~99% of parses with no behavior change. Alternatively only start accumulating once `[DONE]`-adjacent/final chunks are plausible, but the substring pre-filter is the minimal change.

src/program_scheduling/runtime.rs
```diff
@@ -0,0 +1,536 @@
...
+                expected_resume: program.expected_resume,
+                placement: program.placement.clone(),
+                estimated_context_tokens: program.estimated_context_tokens,
+                in_flight_requests: program.in_flight_requests,
+                waiting_requests: self.request_pool.program_len(&program.reference),
+            })
+    }
+
+    /// Snapshot all live Program lifecycle facts without imposing order.
+    pub(crate) fn views(&self) -> Vec<RuntimeProgramView> {
```
wuhang2014: **[warning]** `views()` clones a `ProgramRef` (two heap `String`s) per live program, and it is invoked repeatedly per tracked request — `target_usage`, `binding_candidates`, `rank_admission_plan` (once per candidate), `repair_capacity`, `expire_acting_ttls`, `publish_metrics`, … — all under the single scheduler `Mutex`. Combined with the default `paused_retention_ttl_seconds: 1800`, steady-state program count ≈ program-arrival-rate × 1800, so at high agent churn each arrival/completion becomes an O(P) string-cloning scan per call site (effectively O(P×T) per admission cycle).

Fine at the validated scale (≤64 programs/target), but worth hardening before large deployments: per-target program indexes for the hot filters, `Arc<str>` in `ProgramRef` to make views cheap to snapshot, and/or reusing one `views()` snapshot per decision pass instead of re-collecting per helper.

src/program_scheduling/domain.rs
```diff
@@ -0,0 +1,226 @@
...
+        &self.program_id
+    }
+
+    /// Monotonic generation assigned when this Program is materialized.
+    pub fn generation(&self) -> u64 {
+        self.generation
+    }
+
+    /// Stable privacy-preserving identifier suitable for diagnostics.
+    pub fn redacted_id(&self) -> String {
```
wuhang2014, line 65: **[nit]** `DefaultHasher::new()` is SipHash with fixed zero keys, so `redacted_id()` is an unsalted, reproducible hash — anyone who knows or can guess a `program_id` (they follow predictable `task:agent` patterns) can confirm the mapping, which undercuts the "privacy-preserving" doc comment on diagnostics. Salting with a per-process random key (lazily initialized `OnceLock<u64>` from `rand`/address entropy) would make it non-reversible across router restarts while staying stable within a process.

wuhang2014 (commented):
Independent review pass at a77107f. The previously flagged items (bounded identity contexts, target snapshot races, retry occupancy accounting) all look addressed. Remaining findings below — the first two are cross-cutting behavior changes that apply even when Program scheduling is disabled, which contradicts the PR description's invariant that native routing/forwarding paths remain unchanged.

docs/program_scheduling.md
```diff
@@ -0,0 +1,192 @@
...
+
+Program scheduling is an optional layer before native vLLM Router forwarding. It groups related requests into a long-lived Program, retains requests in a Router RequestPool when a backend has insufficient logical capacity, and protects reusable KV across short tool calls with Progress-TTL. Requests that do not carry the configured Program metadata continue through the existing request-level routing policy.
+
+## Enablement
+
+Library users configure Program scheduling through `RouterConfig.program_scheduling`. The standalone Rust and Python CLIs accept the same object through `--program-scheduling-config-json`; individual scheduling fields are not exposed as separate CLI flags. The `program_scheduling_enable_key` field controls which requests opt in:
+
+| Value | Behavior |
+|---|---|
+| `vllm_xargs.agentic_context` | Default. Only the canonical body object enables Program scheduling. |
```
wuhang2014, line 11: Why this is the default behavior? How many agent use `agentic_context` as carrier?

docs/program_scheduling.md
```diff
@@ -0,0 +1,192 @@
...
+
+vllm-router \
+  --host 0.0.0.0 \
+  --port 3001 \
+  --worker-urls http://worker-0:8000 http://worker-1:8000 \
+  --intra-node-data-parallel-size 2 \
+  --policy consistent_hash \
+  --prometheus-host 0.0.0.0 \
+  --prometheus-port 29000 \
+  --program-scheduling-config-json "${PROGRAM_CONFIG_JSON}"
```
wuhang2014, line 67: Is there a way that enable program cheduling without specifying config file? Since each key in config file has a default value.

docs/program_scheduling.md
```diff
@@ -0,0 +1,192 @@
...
+| `binding_strategy` | `consistent_hash` | Selects the initial home Rank for each new Program generation. Supported values are `consistent_hash`, `program_round_robin`, `least_program_count`, `reasoning_token_balance`, and `cache_aware`; it does not replace admission or resume decisions. |
+| `hash_virtual_nodes` | `160` | Sets the number of virtual nodes used by `consistent_hash` initial binding. It has no effect on the other binding strategies. |
+| `token_capacity_per_target` | `null` | Supplies the logical KV-token capacity of each concrete DP Rank. A value enables token-based admission, growth reserve, capacity repair, and cross-Rank headroom checks; `null` falls back to the Program-count guard. `reasoning_token_balance` requires this value. |
+| `max_active_programs_per_target` | `64` | Limits active Programs on one Rank as a coarse concurrency guard. It is the primary capacity guard when token capacity is absent and remains a count guard when token capacity is configured; forced resume may override it. |
```
wuhang2014, line 97: Is it better to name it `per_dp_rank`?

src/routers/http/router.rs
warriorsniu: Agreed — this changed the transparent-proxy path even when Program scheduling was disabled, and forwarding the client Authorization before adding the Router credential could produce two Authorization values. It also lacked the required hop-by-hop filtering. In [`6dd92b3`](https://github.com/vllm-project/router/commit/6dd92b3cead83016de1e40a21c0d07d3cd324c30) I removed the added client-header forwarding and restored the native transparent-proxy behavior. Program identity is consumed by the Router and does not require forwarding arbitrary client headers to the backend; Router-owned Authorization and the existing tracing path remain unchanged.

src/protocols/spec.rs
warriorsniu: Agreed. The conversation-text extraction was intended only for Program binding and estimation, but reusing `extract_text_for_routing()` also changed the native policy input when the feature was disabled. In [`6dd92b3`](https://github.com/vllm-project/router/commit/6dd92b3cead83016de1e40a21c0d07d3cd324c30) I restored the existing session-id-first native routing contract and introduced a separate Program scheduling text path. Full conversation/tool text is constructed only after both the scheduler and an opted-in Program identity are present; native ConsistentHash/RendezvousHash and CacheAware routing retain their previous inputs. The Program path retains the same conversation text and fallback behavior as before this fix, so its token estimation and CacheAware binding semantics are unchanged.

src/routers/http/program_adapter.rs
warriorsniu: Good catch. This only runs for Program-tracked streams, but parsing every token delta is unnecessary because usage appears only in the usage-bearing chunks. In [`9334c39`](https://github.com/vllm-project/router/commit/9334c3966aea4831e2cff3936b3f9fc6f6a5cf8f) I added a byte-level `"usage"` prefilter before JSON parsing while preserving partial-line buffering, final-line handling, and all supported usage envelopes. The regression test covers usage and non-usage chunks across chunk boundaries.

src/program_scheduling/runtime.rs
warriorsniu: Agreed. In [`3881953`](https://github.com/vllm-project/router/commit/388195329634a2647a72c0a9fafa1f9aa2e1d23a), [`af59141`](https://github.com/vllm-project/router/commit/af591411a120d7ebe88c47dfa85252e87bb8c239), and [`0f14722`](https://github.com/vllm-project/router/commit/0f14722) I removed both forms of the runtime snapshot: the owned `views()` that cloned every live Program's strings, and the remaining borrowed-view `Vec` collections. Observation capture, binding candidates, and metrics now update per-target summaries directly in one runtime pass; diagnostics builds only its final API output. Helpers that mutate runtime retain only exact generation-fenced `ProgramRef` values for actual TTL/release or bounded active victim candidates. I deliberately preserved the existing observation/commit boundaries: TTL and release still freeze candidates before mutation, while capacity repair still observes, sorts, and commits one target at a time, so later targets continue to see earlier committed transitions. The follow-up restores the original commit-time private-token read for projected-usage accounting, while keeping only the ranking key frozen by the scan. Admission, fairness, usage, ordering, and fallback formulas are unchanged. I kept per-target incremental indexes out of this PR because they would add a new consistency obligation to every lifecycle transition. The 72 Program Scheduling tests and the full 610-test library suite pass unchanged.

src/program_scheduling/domain.rs
warriorsniu: Agreed. The previous identifier removed plaintext from logs but was deterministic enough for guessed Program IDs, so the privacy-preserving wording was too strong. In [`ca893ed`](https://github.com/vllm-project/router/commit/ca893edaddfcc4237b92ccdb4d04959ca918c0cd) I switched it to a process-randomized `RandomState` held in `OnceLock`. IDs remain stable within one Router process for log correlation but rotate across restarts, without adding a dependency. Tests cover within-process stability, separation across generations/model pools, and output format.

docs/program_scheduling.md
warriorsniu: Good question. This default is not based on `vllm_xargs.agentic_context` already being emitted by a large number of agents—it currently is not. It is the complete per-request opt-in contract introduced for Program scheduling and is injected today by the AgentInfer adapters, [replay transport](https://github.com/openJiuwen-ai/agent-infer/blob/f764c9a7c387eef632defa96fb3ccae6f654c0f9/agentinfer/agentbench/replay/transport.py#L198-L204), and benchmark tooling.
 The design follows the same general pattern as [Dynamo's `nvext.agent_hints`](https://docs.nvidia.com/dynamo/dev/agents/agent-hints): serving-specific metadata lives in an extensible request namespace and normally requires a serving-aware client or harness adapter. Dynamo's own [agent-harness documentation](https://docs.nvidia.com/dynamo/dev/agents/agent-harnesses) shows that mainstream Claude Code, Codex, and OpenCode integrations primarily emit session headers rather than `nvext.agent_hints` themselves. We use `vllm_xargs.agentic_context` because Program scheduling needs both stable Program/tree identity and request-scoped controls such as priority, deadline, expected output length, expected resume, and KV-retention TTL. The currently supported `agent_hint` compatibility shape only provides session/parent identity and an ephemeral cache marker, so it cannot express the full roadmap contract.
 The [Program-scheduling RFC](https://github.com/vllm-project/router/issues/285) also requires requests without the configured Program metadata to remain on the native request-level routing path. Keeping the strict contract as the default preserves that opt-in boundary and prevents an unrelated session header from unexpectedly enabling Router queueing, Progress-TTL, pause/resume, and capacity accounting. Deployments that explicitly want the broader identity-only compatibility behavior can select `auto`; within `auto`, `vllm_xargs.agentic_context` still has precedence when multiple sources are present. I will clarify the rationale, integration requirement, and strict-versus-`auto` behavior in the documentation.

docs/program_scheduling.md
warriorsniu: Agreed. Using `--program-scheduling-config-json '{}'` as the feature-presence switch overloads a configuration argument and makes the default setup unnecessarily obscure. I added a dedicated `--enable-program-scheduling` switch. The switch alone enables Program scheduling with `ProgramSchedulingConfig::default()` and requires no file; `--program-scheduling-config-json` now only supplies optional overrides and is rejected without the enable switch. The Rust CLI, Python CLI, and Python binding use the same four-state behavior. I will keep the file-based example for calibrated production settings.

docs/program_scheduling.md
warriorsniu: Yes, `per_dp_rank` is clearer at the user-facing configuration boundary. The scheduler's internal `target` abstraction is one concrete worker engine or internal DP rank, but operators size this value as KV-token capacity for each DP rank. I will rename the documented and serialized field to `token_capacity_per_dp_rank` and update validation/tests accordingly; the internal target abstraction can remain unchanged.

wuhang2014 (approved):

PR #287, opened 2026-09-18
fix: advance sequence read offset after append_text

### LINKED ISSUE #258: [Bug]: Sequence::append_text makes the next token repeat existing text
I created an empty `Sequence`, called `append_text("a")`, then appended the token for `"b"`.

Expected: `append_token` returns `"b"`, the newly decoded text.

Actual: it returns `"ab"`, repeating the text supplied to `append_text`.

The property test shrank this to a one-token prefix followed by one token. Tested on `main` commit `ab7e4537654caf8a2096bbe7c65e138754c72d6c` with an offline in-memory tokenizer.

In [src/tokenizer/sequence.rs:84–108](https://github.com/vllm-project/router/blob/ab7e4537654caf8a2096bbe7c65e138754c72d6c/src/tokenizer/sequence.rs#L84-L108), `append_text` extends `token_ids` without advancing `read_offset`. The next `append_token` sees the previous read offset as zero and decodes the entire token list. The method's documentation says it returns newly decoded text.

The included offline tokenizer maps token 0 to `a` and token 1 to `b`. After `append_text("a")`, `append_token(1)` returns `ab` instead of the expected increment `b`. No downloaded tokenizer or model is needed.

### Standalone reproduction

In a checkout of `ab7e4537654caf8a2096bbe7c65e138754c72d6c`, save the code below as `tests/issue_258.rs`, then run:

```sh
cargo test --locked --test issue_258 -- --nocapture
```

<details>

```rust
use vllm_router_rs::tokenizer::sequence::Sequence;
use anyhow::Result;
use vllm_router_rs::tokenizer::traits::{Decoder, Encoder, Encoding, SpecialTokens, Tokenizer};

const PIECES: [&str; 8] = ["a", "b", " ", "é", "🙂", "<special>", "x", "\n"];

struct PieceTokenizer {
    special_tokens: SpecialTokens,
    piece_to_id: HashMap<&'static str, u32>,
}

impl PieceTokenizer {
    fn new() -> Self {
        let piece_to_id = PIECES
            .iter()
            .enumerate()
            .map(|(id, piece)| (*piece, id as u32))
            .collect();
        Self {
            special_tokens: SpecialTokens {
                bos_token: None,
                eos_token: Some("<special>".to_string()),
                unk_token: None,
                sep_token: None,
                pad_token: None,
                cls_token: None,
                mask_token: None,
                additional_special_tokens: Vec::new(),
            },
            piece_to_id,
        }
    }

    fn expected_decode(ids: &[u32], skip_special_tokens: bool) -> String {
        ids.iter()
            .filter_map(|id| PIECES.get(*id as usize))
            .filter(|piece| !(skip_special_tokens && **piece == "<special>"))
            .copied()
            .collect()
    }
}

impl Encoder for PieceTokenizer {
    fn encode(&self, input: &str) -> Result<Encoding> {
        let mut ids = Vec::new();
        let mut rest = input;
        while !rest.is_empty() {
            let Some(piece) = PIECES
                .iter()
                .filter(|piece| rest.starts_with(**piece))
                .max_by_key(|piece| piece.len())
            else {
                anyhow::bail!("input has no piece in the offline test vocabulary: {rest:?}");
            };
            ids.push(self.piece_to_id[piece]);
            rest = &rest[piece.len()..];
        }
        Ok(Encoding::Sp(ids))
    }

    fn encode_batch(&self, inputs: &[&str]) -> Result<Vec<Encoding>> {
        inputs.iter().map(|input| self.encode(input)).collect()
    }
}

impl Decoder for PieceTokenizer {
    fn decode(&self, token_ids: &[u32], skip_special_tokens: bool) -> Result<String> {
        Ok(Self::expected_decode(token_ids, skip_special_tokens))
    }
}

impl Tokenizer for PieceTokenizer {
    fn vocab_size(&self) -> usize {
        PIECES.len()
    }

    fn get_special_tokens(&self) -> &SpecialTokens {
        &self.special_tokens
    }

    fn token_to_id(&self, token: &str) -> Option<u32> {
        self.piece_to_id.get(token).copied()
    }

    fn id_to_token(&self, id: u32) -> Option<String> {
        PIECES.get(id as usize).map(|piece| (*piece).to_string())
    }
}

#[test]
fn known_failure_append_text_then_token_repro() {
    let tokenizer = Arc::new(PieceTokenizer::new());
    let mut sequence = Sequence::new(tokenizer);
    sequence.append_text("a").unwrap();

    assert_eq!(sequence.append_token(1).unwrap(), "b");
}
```

</details>

### Verified failure output

```text
tests/issue_258.rs:97:5
  left: "ab"
 right: "b"
```

DatenpflegeNordHL:
Summary

- advance "read_offset" after "append_text()" extends the sequence token IDs
- prevent the next "append_token()" call from returning text that was already appended
- add focused regression coverage for the incremental decoding behavior

Testing

- "cargo test --lib tokenizer::sequence::tests::test_sequence_append_token_after_text -- --exact"
- "cargo test --lib tokenizer::sequence::tests"
- "cargo fmt --check"
- "cargo test --lib"

Fixes #258


hsliuustc0106 (commented):
Reviewed at `1eb75751`. Correct minimal fix: `append_text` extended `token_ids` without advancing `read_offset`, so a following `append_token` decoded an empty range and returned "" instead of the delta (the new `test_sequence_append_token_after_text` reproduces exactly that: `append_text("Hello")` then `append_token(2)` must yield `" world"`). The only other `append_text` callers are tests, so no other expectations change. No findings.

---
*From an automated daily review pass over new/updated PRs (head SHA frozen at `1eb75751`).*

PR #288, opened 2026-09-18
fix: count Unicode characters in logprob offsets

### LINKED ISSUE #263: [Bug]: PD logprobs use byte offsets for non-ASCII prompt tokens
I merged a completion response whose last prompt token was `é` at offset 0 with a decode token at offset 0.

Expected: the first decode token has `text_offset = 1`, matching vLLM's character-count offsets.

Actual: the merged response has `text_offset = 2`.

ASCII tokens are unaffected. The property test also reproduces this with CJK and emoji tokens. Tested on `main` commit `ab7e4537654caf8a2096bbe7c65e138754c72d6c` through the logprobs merge helper. I did not run a live prefill/decode deployment.

In [src/routers/http/logprobs_merge.rs:189–198](https://github.com/vllm-project/router/blob/ab7e4537654caf8a2096bbe7c65e138754c72d6c/src/routers/http/logprobs_merge.rs#L189-L198), the merge base is the last prompt offset plus Rust `str::len()`, which counts UTF-8 bytes. The upstream [vLLM completion implementation](https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/completion/serving.py) advances offsets using Python `len(token)`, which counts Unicode code points.

For `é`, these lengths are 2 and 1 respectively. The complete input objects are included below: one prompt token at offset 0, plus one decode token at offset 0. Expected merged `text_offset`: `[0, 1]`. Actual: `[0, 2]`. The extra prefill entry named `discarded` represents the prefill-generated token removed by this helper; `prompt_logprobs` has one entry, identifying one prompt token. This demonstrates the merge arithmetic with constructed JSON, not a captured live response.

### Standalone reproduction

In a checkout of `ab7e4537654caf8a2096bbe7c65e138754c72d6c`, save the code below as `tests/issue_263.rs`, then run:

```sh
cargo test --locked --test issue_263 -- --nocapture
```

<details>

```rust
use serde_json::json;
use vllm_router_rs::routers::http::logprobs_merge::merge_logprobs_in_json;

#[test]
fn unicode_prompt_logprobs_offset_repro() {
    let prefill = json!({"choices": [{
        "prompt_logprobs": [null],
        "logprobs": {
            "token_logprobs": [null, -1],
            "tokens": ["é", "discarded"],
            "text_offset": [0, 999],
            "top_logprobs": [null, null]
        }
    }]});
    let mut decode = json!({"choices": [{"logprobs": {
        "token_logprobs": [-2],
        "tokens": ["decode"],
        "text_offset": [0],
        "top_logprobs": [null]
    }}]});

    assert!(merge_logprobs_in_json(&prefill, &mut decode));
    assert_eq!(
        decode["choices"][0]["logprobs"]["text_offset"],
        json!([0, 1])
    );
}
```

</details>

### Verified failure output

```text
tests/issue_263.rs:23:5
  left: Array [Number(0), Number(2)]
 right: Array [Number(0), Number(1)]
```

DatenpflegeNordHL:
Summary

- count Unicode characters instead of UTF-8 bytes when advancing merged logprob offsets
- keep non-ASCII offsets consistent with vLLM's character-based semantics
- add regression coverage for ASCII, accented, CJK, and emoji tokens

Testing

- "cargo test --lib routers::http::logprobs_merge::tests::test_merge_logprobs_uses_character_offsets -- --exact"
- "cargo test --lib routers::http::logprobs_merge::tests"
- "cargo fmt --check"
- "cargo test --lib"

Fixes #263


hsliuustc0106 (commented):
Reviewed at `be0b961a`. Correct fix: the merged `text_offset` continuation counted the last prefill prompt token in bytes (`s.len()`), so multi-byte tokens (é, 界, 🙂) shifted all decode offsets. `chars().count()` matches the code-point semantics of the upstream `text_offset` array, and the test matrix covers ASCII through 4-byte emoji. No findings.

---
*From an automated daily review pass over new/updated PRs (head SHA frozen at `be0b961a`).*

PR #289, opened 2026-09-18
fix: clean worker indexes on re-registration

### LINKED ISSUE #255: [Bug]: re-registering a worker URL leaves stale registry indexes
I registered a worker URL, then registered the same URL again with a different model and worker type.

After the second registration, `get_all()` contains one worker, but the old model and worker-type indexes still return the URL. Re-registering the same value also creates duplicate entries in the connection index.

Expected: registering an existing URL replaces that worker in the primary map and every secondary index.

Actual: the primary entry is replaced while duplicate and stale secondary-index entries remain.

Tested on `main` commit `ab7e4537654caf8a2096bbe7c65e138754c72d6c` with a single-threaded component test.

In [src/core/worker_registry.rs:76–120](https://github.com/vllm-project/router/blob/ab7e4537654caf8a2096bbe7c65e138754c72d6c/src/core/worker_registry.rs#L76-L120), registering an existing URL reuses its worker ID and replaces the primary entry, then appends that ID to the model, type, and connection indexes. The shown path does not first remove the previous index entries.

The test uses the same URL `http://worker-0`, first with label `model_id=model-0` and type `Regular`, then with `model_id=model-1` and type `Decode`. The first failing assertion is that `get_by_model("model-0")` should be empty. The expected replacement behavior is also described by the existing-URL comment in `register`.

### Standalone reproduction

In a checkout of `ab7e4537654caf8a2096bbe7c65e138754c72d6c`, save the code below as `tests/issue_255.rs`, then run:

```sh
cargo test --locked --test issue_255 -- --nocapture
```

<details>

```rust
use vllm_router_rs::core::{BasicWorker, Worker, WorkerRegistry, WorkerType};

fn worker(url_slot: u8, model_slot: u8, kind_slot: u8) -> Arc<dyn Worker> {
    let mut labels = HashMap::new();
    labels.insert("model_id".to_string(), format!("model-{}", model_slot % 3));
    let kind = match kind_slot % 3 {
        0 => WorkerType::Regular,
        1 => WorkerType::Prefill {
            bootstrap_port: None,
        },
        _ => WorkerType::Decode,
    };
    Arc::new(BasicWorker::new(format!("http://worker-{url_slot}"), kind).with_labels(labels))
}

#[test]
fn worker_registry_duplicate_url_repro() {
    let registry = WorkerRegistry::new();
    registry.register(worker(0, 0, 0));
    registry.register(worker(0, 1, 2));

    assert_eq!(registry.get_all().len(), 1);
    assert!(registry.get_by_model("model-0").is_empty());
    assert_eq!(registry.get_by_model("model-1").len(), 1);
    assert_eq!(registry.get_by_model_fast("model-1").len(), 1);
    assert_eq!(registry.get_by_type(&WorkerType::Regular).len(), 0);
    assert_eq!(registry.get_by_type(&WorkerType::Decode).len(), 1);
}
```

</details>

### Verified failure output

```text
tests/issue_255.rs:24:5
assertion failed: registry.get_by_model("model-0").is_empty()
```

DatenpflegeNordHL:
Summary

- clean the previous worker's secondary indexes before replacing a worker registered with the same URL
- preserve the existing worker ID while rebuilding model, type, and connection indexes
- serialize registry mutations so cleanup and re-registration cannot interleave
- add regression coverage for stale and duplicate secondary-index entries

Testing

- "cargo test --lib core::worker_registry::tests::test_reregister_worker_cleans_secondary_indexes -- --exact"
- "cargo test --lib core::worker_registry::tests"
- "cargo fmt --check"
- "cargo test --lib"

Fixes #255


hsliuustc0106 (commented):
Reviewed at `64caf081`. Correct fix: previously, re-registering a worker by URL (e.g. after a model/type/label change) left stale entries in `type_workers`, `connection_workers`, and the model indexes. `register()` now purges all secondary indexes via `remove_inner` while preserving the WorkerId, and the new `mutation_lock` serializes the multi-map mutations (no lock nesting — public `remove`/`remove_by_url` route through the same inner path). The test covering label change + type change + removal is thorough. No findings.

---
*From an automated daily review pass over new/updated PRs (head SHA frozen at `64caf081`).*

PR #292, opened 2026-09-19
fix: keep response user index consistent on overwrite

### LINKED ISSUE #256: [Bug]: overwriting a response ID leaves it in the old user's index
I stored a response ID for `user-0`, then stored the same response ID for `user-1`.

Expected: replacing a response ID removes it from the previous user's index and adds one entry for the new user.

Actual: `user-0` still lists the response after it has been replaced with `user-1`'s response. Storing the same ID repeatedly for one user also creates duplicate list entries.

Tested on `main` commit `ab7e4537654caf8a2096bbe7c65e138754c72d6c`. This is a direct in-memory storage API reproduction. I have not shown that untrusted clients can choose colliding response IDs through the HTTP API.

In [src/data_connector/response_memory_store.rs:60–78](https://github.com/vllm-project/router/blob/ab7e4537654caf8a2096bbe7c65e138754c72d6c/src/data_connector/response_memory_store.rs#L60-L78), `store_response` appends the ID to the new user's index and replaces the response map entry. It does not remove the ID from the old user's list. [src/data_connector/response_memory_store.rs:146–171](https://github.com/vllm-project/router/blob/ab7e4537654caf8a2096bbe7c65e138754c72d6c/src/data_connector/response_memory_store.rs#L146-L171) then resolves those indexed IDs against the current response map.

Consequently, the direct storage API can list the replacement response under the previous user. The test uses ID `response-0`, stores it for `user-0`, then overwrites it for `user-1`, and expects `user-0`'s list to be empty. This is an index-consistency report; HTTP exploitability has not been established.

### Standalone reproduction

In a checkout of `ab7e4537654caf8a2096bbe7c65e138754c72d6c`, save the code below as `tests/issue_256.rs`, then run:

```sh
cargo test --locked --test issue_256 -- --nocapture
```

<details>

```rust
use chrono::{TimeZone, Utc};
use vllm_router_rs::data_connector::{MemoryResponseStorage, ResponseId, ResponseStorage, StoredResponse};

fn response(id: u8, user: u8, sequence: usize) -> StoredResponse {
    let mut response = StoredResponse::new(
        format!("input-{id}-{sequence}"),
        format!("output-{id}-{sequence}"),
        None,
    );
    response.id = ResponseId::from_string(format!("response-{id}"));
    response.user = Some(format!("user-{user}"));
    response.created_at = Utc
        .timestamp_opt(1_700_000_000 + sequence as i64, 0)
        .unwrap();
    response
}

#[test]
fn response_store_cross_user_overwrite_repro() {
    let runtime = tokio::runtime::Builder::new_current_thread()
        .enable_all()
        .build()
        .unwrap();
    let store = MemoryResponseStorage::new();

    runtime
        .block_on(store.store_response(response(0, 0, 0)))
        .unwrap();
    runtime
        .block_on(store.store_response(response(0, 1, 1)))
        .unwrap();

    assert!(runtime
        .block_on(store.list_user_responses("user-0", None))
        .unwrap()
        .is_empty());
    assert_eq!(
        runtime
            .block_on(store.list_user_responses("user-1", None))
            .unwrap()
            .len(),
        1
    );
}
```

</details>

### Verified failure output

```text
tests/issue_256.rs:33:5
assertion failed: runtime.block_on(store.list_user_responses("user-0",
                None)).unwrap().is_empty()
```

RainieLLM:
Fixes #256

`MemoryResponseStorage::store_response` could leave a response ID in the previous user's index when the ID was overwritten. Reusing the same ID for the same user also added duplicate index entries.

This change removes all existing index entries for the old user before the response is replaced, then adds the ID to the new user's index once. The response map and user index remain consistent under the existing write lock.

Added regression tests for cross-user and same-user overwrites.


RainieLLM:
I've added the missing Signed-off-by line to the commit. The code is unchanged. Could a maintainer approve the CI workflows and take a look when you have time? Thanks for your help!

